PR #19, opened 2019-06-28
Add --with-pkgconfigdir option

barrbrain:
Useful for odd distributions like Homebrew. In my case, `/usr/local/lib/pkgconfig` was the path.
A default of `/usr/local/lib/pkg-config` was inferred by cargo-c instead.
For consistency with autotools[1] the option is named `--with-pkgconfigdir`.

[1] https://www.mankier.com/7/pkg.m4#Autoconf_Macros-Pkg_installdir(Directory)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,6 +36,8 @@ struct Common {
     includedir: Option<PathBuf>,
     #[structopt(long = "bindir", parse(from_os_str))]
     bindir: Option<PathBuf>,
+    #[structopt(long = "with-pkgconfigdir", parse(from_os_str))]
+    pkgconfigdir: Option<PathBuf>,
 
     /// Space-separated list of features to activate
     #[structopt(long = "features")]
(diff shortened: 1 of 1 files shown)


dwbuiten:
I'm not sure this is needed. Shouldn't it always be `pkgconfig`?

lu-zero:
It should be `pkgconfig`, you spotted another bug :)

The option itself is a good idea.

src/main.rs
```diff
@@ -248,6 +251,7 @@ impl Config {
         let prefix = opt.prefix.unwrap_or_else(|| "/usr/local".into());
         let libdir = opt.libdir.unwrap_or_else(|| prefix.join("lib"));
         let includedir = opt.includedir.unwrap_or_else(|| prefix.join("include"));
+        let pkgconfigdir = opt.pkgconfigdir.unwrap_or_else(|| libdir.join("pkg-config"));
```
lu-zero:
```suggestion
        let pkgconfigdir = opt.pkgconfigdir.unwrap_or_else(|| libdir.join("pkgconfig"));
```

barrbrain:
My `man 1 pkg-config` says:
> By default the directory is $libdir/pkg-config, but the default can be changed by passing DIRECTORY.

So, perhaps it is different in some distributions.

Issue #27, opened 2019-07-09
Support for cargo --jobs


wiiaboo:
Would be cool to support `--jobs` so I can use my PC while building stuff with cargo-c.

Not sure if I should open another issue for this, but being able to decide whether to build the dynamic and/or static libs would be cool too.

wiiaboo:
Also, this is a low priority feature request. Meanwhile, I'm good with using $CARGO_HOME/config with 
```ini
[build]
jobs = 1
```

lu-zero:
If you want to try to send a patch I'd land this before the next release.

wiiaboo:
Might take a while, as I have zero experience with rust.

wiiaboo:
This seems to work, but I feel like I'm doing something wrong with the `cmd.arg("--jobs").arg(&format!("{}", self.jobs.unwrap()));` line. Is this sane?

```diff
diff --git a/src/main.rs b/src/main.rs
index 6f22c8d..bc32122 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,9 @@ struct Common {
     /// Path to the project, by default the current working directory
     #[structopt(long = "project-dir", parse(from_os_str))]
     projectdir: Option<PathBuf>,
+    /// Number of rustc jobs to run in parallel.
+    #[structopt(long = "jobs")]
+    jobs: Option<u32>,
     /// Build artifacts in release mode, with optimizations
     #[structopt(long = "release")]
     release: bool,
@@ -194,6 +197,7 @@ struct BuildInfo {
 struct Config {
     /// The library name
     name: String,
+    jobs: Option<u32>,
     /// Build artifacts in release mode, with optimizations
     release: bool,
     /// Build for the target triple or the host system.
@@ -265,6 +269,7 @@ impl Config {
 
         Config {
             name,
+            jobs: opt.jobs,
             release: opt.release,
             target: Target::new(opt.target.as_ref()).unwrap(),
             destdir: opt.destdir.unwrap_or_else(|| PathBuf::from("/")),
@@ -420,6 +425,10 @@ impl Config {
         cmd.arg("--target-dir").arg(&self.target_dir);
         cmd.arg("--manifest-path").arg(&self.pkg.manifest_path);
 
+        if self.jobs.is_some() {
+            cmd.arg("--jobs").arg(&format!("{}", self.jobs.unwrap()));
+        }
+
         if let Some(t) = self.target.verbatim.as_ref() {
[...]

PR #29, opened 2019-07-11
cli: pass --jobs to cargo

wiiaboo:
c.f. #27 

The `.into()` didn't seem to work due to E0282, I didn't investigate too hard on how to fix that as it's already a bit late.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,9 @@ struct Common {
     /// Path to the project, by default the current working directory
     #[structopt(long = "project-dir", parse(from_os_str))]
     projectdir: Option<PathBuf>,
+    /// Number of rustc jobs to run in parallel.
+    #[structopt(long = "jobs")]
+    jobs: Option<u32>,
     /// Build artifacts in release mode, with optimizations
     #[structopt(long = "release")]
     release: bool,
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -434,6 +437,10 @@ impl Config {
         cmd.arg("--target-dir").arg(&self.target_dir);
         cmd.arg("--manifest-path").arg(&self.pkg.manifest_path);
 
+        if let Some(jobs) = self.cli.jobs {
```
lu-zero:
```suggestion
        if let Some(jobs) = self.cli.jobs.as_ref() {
```

src/main.rs
```diff
@@ -434,6 +437,10 @@ impl Config {
...
+            cmd.arg("--jobs").arg(jobs.to_string());
```
lu-zero:
```suggestion
            cmd.arg("--jobs").arg(jobs);
```

wiiaboo: `error[E0277]: the trait bound `u32: std::convert::AsRef<std::ffi::OsStr>` is not satisfied`
I've tried that before. Only works if I change `jobs: Option<u32>` to `jobs: Option<String>`

lu-zero: You are right :)

PR #30, opened 2019-07-22
Travis: Add basic testing and coverage

### LINKED ISSUE #15: Add some minimal CI
- [ ] Setup Travis
- [ ] Setup kcov

1480c1:
Closes #15

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
     - rust: nightly
   fast_finish: true
 before_install:
+  - "sudo chown -R travis: $HOME/.ccache"
   - export PATH="/usr/local/opt/ccache/libexec:/usr/lib/ccache:$PATH"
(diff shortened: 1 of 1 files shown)


1480c1:
Still making sure it's correct.

To actually activate coveralls.io, you would have to go to <https://coveralls.io> and add repo

lu-zero:
Thank you :)

lu-zero:
coveralls times out on install apparently.

1480c1:
Trying to figure out what's wrong

1480c1:
Decided to manually install kcov instead of using the script since that seemed to not always work. Should be fixed now

1480c1:
Would it be possible to add more tests? particularly for static_libs.rs and main.rs

lu-zero:
I'd like to have an integration test so we are sure cargo-c keeps producing the same files, but might be a bit more involving.

1480c1:
That is true

lu-zero:
Let's land this first and then add more tests :)

PR #38, opened 2019-11-04
[WIP] version header

tdaede:


lu-zero:
The missing parts:
- [ ] Add a key either in the main `Cargo.toml` or in a stand alone toml pass the prefix wanted and if the version information should be produced as macro symbols in the `.h`
- [ ] Use the version information already extracted from `Cargo.toml` to produce the actual symbols.

lu-zero:
It does not work yet since the macros appear before the include guard.

Issue #44, opened 2019-11-26
Request: provide the Cargo.lock file


dbermond:
Please consider providing the `Cargo.lock` file. This file is important for making [reproducible builds](https://reproducible-builds.org/).

Without the lockfile, a given version of this project cannot be reproducible over time, as the dependencies will get updated. Using a lockfile with the `--locked` option in `cargo` commands (like `cargo build`) will ensure that the dependencies will not be updated, allowing the build to be reproducible. It can also prevent possible build problems in updated dependency versions.

A quote from the [`cargo install` documentation](https://doc.rust-lang.org/cargo/commands/cargo-install.html) better explains the importance of the lockfile:

> The --locked flag can be used to force Cargo to use the packaged Cargo.lock file if it is available. This may be useful for ensuring reproducible builds, to use the exact same set of dependencies that were available when the package was published. It may also be useful if a newer version of a dependency is published that no longer builds on your system, or has other problems.

lu-zero:
Who cares about reproducible builds should store the `cargo vendor` output.

I consider the Cargo.lock purely annoying in the git tree.

I would accept patches that save the `Cargo.lock` as release artifact either by generating the tarballs manually or by just adding it as-is.

PR #60, opened 2020-03-23
Add cargo.lock file

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/issues/46025

It would be nice to have a cargo.lock file to enforce reproducible build.


.github/workflows/rust.yml
```diff
@@ -72,12 +72,12 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --verbose
+          args: --locked --verbose
```
lu-zero: I'd rather have a separate set of CI targets for that.

.gitignore
```diff
@@ -1,3 +1,2 @@
 /target
 **/*.rs.bk
-Cargo.lock
```
lu-zero: I'd rather have a CI action to add the lock to the release. @Luni-4 can you bake the same magic you put on rav1e please? :)

lu-zero:
Please refer to https://github.com/lu-zero/cargo-c/issues/44#issuecomment-558784195

Having Cargo.lock in the git tree is highly disruptive

Luni-4:
@chenrui333 

You could create a `deploy.yml` script that does the following steps:
- Run `cargo update` to generate the `Cargo.lock` file
- Update `Cargo.lock` alongside with the release, look at [here](https://github.com/xiph/rav1e/blob/master/.github/workflows/deploy.yml#L97)

The `deploy.yml` file runs when a new tag will be created. Look at [here](https://github.com/xiph/rav1e/blob/master/.github/workflows/deploy.yml#L3) to set tags for `GitHub Actions`. In this case we need only the `'v*.*.*'` pattern.

If you need some help, ping me :)

anthraxx:
Is this issue not resolved with #66? That would be great to have reproducible builds enabled easily.

lu-zero:
Depends on what you use. The `.crate` package provided by crates.io already contains a Cargo.lock (that's what I'm using in Gentoo). The next release will provide the same `Cargo.lock` as release item in case you cannot use the `.crate` package.

Luni-4:
The `Cargo.lock` file is added to a new release, so we can close this one

PR #69, opened 2020-04-24
Use Vec::extend instead of a side-effect-only Option::map

cuviper:


cuviper:
I cited this as an example in https://github.com/rust-lang/rust/issues/71484, which is asking for `#[must_use]` on `Option::map`.

cuviper:
Another possibility could avoid the `Vec`, if you like:

```rust
    let paths = iter::once(&build_targets.include)
        .chain(&build_targets.static_lib)
        .chain(&build_targets.shared_lib);
```

cuviper:
There's a little magic in leveraging `IntoIterator for &Option<T>` here.

lu-zero:
yes, thus why it is a bit surprising, but overall looks compact and I like the expressiveness.

PR #71, opened 2020-04-29
Add msvc target

### LINKED ISSUE #61: msvc support
I get the error message that windows-msvc is not supported. Is it possible to add support for msvc?

Luni-4:
This PR closes #61 


src/cinstall.rs
```diff
@@ -111,7 +111,9 @@ fn cinstall(
 
     if let Some(ref static_lib) = build_targets.static_lib {
         ws.config().shell().status("Installing", "static library")?;
-        fs::copy(static_lib, install_path_lib.join(&format!("lib{}.a", name)))?;
+        if env != "msvc" {
```
lu-zero, line 114: there isn't a static`.lib` to copy here?

dwbuiten:
No, because I neither use MSVC, nor do I have a setup for it.

Luni-4:
I've already tested this PR with `msvc` and it seems to work correctly

src/build.rs
```diff
@@ -86,7 +87,63 @@ fn patch_lib_kind_in_target(ws: &mut Workspace, libkinds: &[&str]) -> anyhow::Re
...
+        dumpbin.arg(format!("/OUT:{}", txt_path.to_str().unwrap()));
+
+        let out = dumpbin.output()?;
+        if out.status.success() {
+            let txt_file = File::open(txt_path)?;
+            let buf_reader = BufReader::new(txt_file);
+            let mut def_file = File::create(targetdir.join(format!("{}.def", name)))?;
+            writeln!(def_file, "{}", "EXPORTS".to_string())?;
+
+            // Skip 16 lines and, for each line, delete all the characters
```
dwbuiten, line 119: Is this true for all supported versions of `dumpbin`? Same for below. It seems possibly a little brittle to hardcode numbers of lines or spacs.

Luni-4: I've modified that code starting from the following loop:

`for /f "skip=19 tokens=4" %A in (file.txt) do echo %A >> file.def`

This loop skips 19 lines because `dumpbin` adds 3 lines of copyright information that are printed on the shell when we use the `/OUT:file.txt' option. Searching through the net I've always found that loop above, so we could suppose that is true for all supported versions of `dumpbin`.

src/build.rs
```diff
@@ -86,7 +87,71 @@ fn patch_lib_kind_in_target(ws: &mut Workspace, libkinds: &[&str]) -> anyhow::Re
...
+            // The Rust loop below is analogue to the following loop.
+            // for /f "skip=19 tokens=4" %A in (file.txt) do echo %A > file.def
+            // The most recent versions of dumpbin adds three lines of copyright
+            // information before the relevant content.
+            // If the "/OUT:file.txt" dumpbin's option is used, the three
+            // copyright lines are added to the shell, so the txt file
+            // contains three lines less.
+            // The Rust loop first skips 16 lines and then, for each line,
+            // deletes all the characters up to the fourth space included
+            // (skip=16 tokens=4)
```
dwbuiten, line 128: Thanks.

I just wanted to know if the output syntax/text is stable between versions, i.e. if a regex-based solution is more "safe" or not. If you're satisfied this approach is stable enough, I have no objections.

Luni-4: Sincerely I've thought of a regex-based approach, but it would have complicated more the code in my opinion. So, after having looked around a bit, I've discovered that the current approach is the most used for `.def` files and that's why I've decided to use that, in this way we are more compliant with the most used implementations.

PR #72, opened 2020-05-03
Deploy cargo-c binaries

### LINKED ISSUE #70: Add a CI deploy target
- [ ] add macOS builds as the release artifacts
- [ ] add Windows builds as the release artifacts
- [ ] add Linux builds as the release artifacts

Luni-4:
This PR fixes #70

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,12 +7,121 @@ on:
 
 jobs:
 
+  windows-binaries:
+
+    strategy:
+      matrix:
+        conf:
+         - msvc
+         - gnu
+        include:
+         - conf: msvc
+           toolchain: stable
+         - conf: gnu
+           toolchain: stable-x86_64-pc-windows-gnu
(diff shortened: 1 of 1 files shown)


dwbuiten:
Which windows toolchain does this use? We should arguably provide both, because e.g. most users of some projects like Rav1e would want to build with the GNU ABI, since its main use is in FFmpeg (and mostly all FFmpeg on windows is MinGW).

lu-zero:
> Which windows toolchain does this use? We should arguably provide both, because e.g. most users of some projects like Rav1e would want to build with the GNU ABI, since its main use is in FFmpeg (and mostly all FFmpeg on windows is MinGW).

The default toolchain is a local configuration and should not depend on how the tool is built IIRC.

Should be tested though.

Luni-4:
I've added both the Windows toolchains for now

lu-zero:
you should rebase it I guess.

PR #76, opened 2020-05-06
Deploy on linux using musl

Luni-4:
This PR allows to deploy `Linux` binaries without external dependencies.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -49,21 +49,59 @@ jobs:
         name: cargo-c-windows-${{ matrix.conf }}-binaries
         path: cargo-c-windows-${{ matrix.conf }}.zip
 
+  linux-binaries:
 
-  unix-binaries:
+    runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        conf:
-         - linux
-         - macos
-        include:
-         - conf: linux
-           os: ubuntu-latest
-         - conf: macos
-           os: macos-latest
+    steps:
+    - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


dwbuiten:
Is there a way we can verify this works as intended before merging / releasing?

Luni-4:
Yep, you can try the binaries [here](https://github.com/Luni-4/cargo-c/releases/tag/v0.4.9)

.github/workflows/deploy.yml
```diff
@@ -49,21 +49,59 @@ jobs:
...
+      run: |
+        export OPENSSL_LIB_DIR=$GITHUB_WORKSPACE/lib
+        export OPENSSL_INCLUDE_DIR=$GITHUB_WORKSPACE/include
+        cargo build --target x86_64-unknown-linux-musl --release
+
+    - name: Create zip
+      run: |
+        cd target/x86_64-unknown-linux-musl/release
+        strip cargo-cbuild cargo-cinstall
+        zip $GITHUB_WORKSPACE/cargo-c-linux.zip \
```
lu-zero, line 92: make a tar please :)

Luni-4: Is there a particular reason not to use `zip`?

1480c1: tar is better understood on *nix systems and sometimes they do not come with any un-zip programs by default

Luni-4: Seems a good reason, thank you @1480c1 :)

dwbuiten:
Confirmed them, thanks!

PR #77, opened 2020-05-11
Strip windows binaries directly on Windows

Luni-4:
Simplify the code stripping windows binaries directly on `Windows`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,12 +36,17 @@ jobs:
       run: |
         cargo build --release
 
+    - name: Strip gnu binaries
+      if: matrix.conf == 'gnu'
+      run: |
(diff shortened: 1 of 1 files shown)


dwbuiten:
Actually, is it even valid to strip MSVC binaries? I didn't think it was... but maybe I am misremembering.

Luni-4:
They should be stored in the pdb file, so actually it could be useless. But I suppose we are not doing something wrong running the `strip` command on them the same.

Luni-4:
@dwbuiten 
Do not strip `msvc` binaries

Luni-4:
> Thanks. The reason I asked is because I have vague memories of `strip` causing malformed binaries when using it on MSVC-built things. I don't know if it is still true or not, but no reason to not play it safe.
> 
> LGTM.

Thanks to  you for the review! :)

PR #88, opened 2020-07-01
Add dlltool argument

### LINKED ISSUE #87: Option to set dlltool command
It would be nice if the dlltool command for building the implib could be specified via an option or environment variable (possibly both). Currently it's hardcoded to "dlltool" which does not exist in my build environment since it's mingw-w64 split up for 32 Bit and 64 Bit so dlltool is prefixed by the target triple -> i686-w64-mingw32-dlltool and x86_64-w64-mingw32-dlltool.

k1ll:
Get dlltool command from environment variable DLLTOOL or command line argument --dlltool. If not specified default to "dlltool".

Fixes #87

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -157,6 +157,7 @@ fn build_implib_file(
     name: &str,
     target: &target::Target,
     targetdir: &PathBuf,
+    dlltool: &PathBuf,
 ) -> anyhow::Result<()> {
     let os = &target.os;
     let env = &target.env;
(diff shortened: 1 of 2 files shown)


src/build.rs
```diff
@@ -322,7 +323,22 @@ pub fn cbuild(
 
     if cur_hash.is_none() || prev_hash != cur_hash {
         build_def_file(&ws, &name, &rustc_target, &root_output)?;
-        build_implib_file(&ws, &name, &rustc_target, &root_output)?;
+
+        let mut dlltool = PathBuf::from("dlltool");
```
lu-zero, line 327: this can be
``` rust
let dlltool = std::env::var_os("DLLTOOL").unwrap_or_else(|| PathBuf::from("dlltool"))
```

src/cli.rs
```diff
@@ -19,6 +19,8 @@ struct Common {
     bindir: Option<PathBuf>,
     #[structopt(long = "pkgconfigdir", parse(from_os_str))]
     pkgconfigdir: Option<PathBuf>,
+    #[structopt(long = "dlltool", parse(from_os_str))]
```
lu-zero, line 22: Add a `///` comment along the lines of
```
/// Use the provided dlltool when building for the windows-gnu targets.
```

src/build.rs
```diff
@@ -325,11 +325,7 @@ pub fn cbuild(
     if cur_hash.is_none() || prev_hash != cur_hash {
         build_def_file(&ws, &name, &rustc_target, &root_output)?;
 
-        let mut dlltool = PathBuf::from("dlltool");
-
-        if std::env::var_os("DLLTOOL").is_some() {
-            dlltool = PathBuf::from(std::env::var_os("DLLTOOL").unwrap());
-        }
+        let mut dlltool = std::env::var_os("DLLTOOL").unwrap_or_else(|| PathBuf::from("dlltool"));
```
lu-zero, line 328: It needs a `.map(PathBuf::from)` as well, sorry for omitting it.

src/build.rs
k1ll: Nope did not work. Type mismatch. Any idea how to fix that in the one liner or should i just revert it to the old version?

src/build.rs
k1ll: ok will add it

lu-zero:
Why did you close it?

k1ll:
Sorry haven't used github in a while thought that might request to merge it

lu-zero:
Now it is merged, thank you a lot :)

PR #90, opened 2020-07-08
append_to_destdir: handle Windows paths properly

### LINKED ISSUE #89: cinstall: destdir path joining problems on Windows with absolute paths
I run 

```
cargo cinstall --release --frozen --library-type=cdylib --destdir=C:/foo
```

but things get installed into C:/usr:

```
C:/usr/
└── local
    ├── bin
    │   └── rav1e.dll
    ├── include
    │   └── rav1e
    │       └── rav1e.h
    └── lib
        ├── librav1e.dll.a
        ├── pkgconfig
        │   └── rav1e.pc
        └── rav1e.def
```

lazka:
It currently uses is_absolute() to strip one leading component which doesn't
work for paths like "C:/foo", "C:foo" and "/foo" which are either not absolute
or have multiple components that need to be stripped.

The goal is to join two paths and removing any components from the second
one that would result in the first one being replaced/changed.

push()/join() alters the first path if it starts with a RootDir, Prefix, or both, so
we push everything from the second path except for those.

Fixes #89

### DIFF
--- a/src/cinstall.rs
+++ b/src/cinstall.rs
@@ -1,4 +1,4 @@
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
 use cargo::core::Workspace;
 use cargo::util::command_prelude::opt;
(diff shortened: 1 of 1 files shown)


lu-zero:
Thank you!

PR #95, opened 2020-07-17
Prepare for the next cargo

lu-zero:
It depends on https://github.com/rust-lang/cargo/pull/8494


Luni-4:
Already implemented in e1b15b7

PR #102, opened 2020-08-20
Ping cbindgen patch version to 0.14.3

rzumer:
0.14.4 introduces a breaking change,
so an update of the patch version is undesirable
for stability reasons.


1480c1:
Do you mean pin, or ping?

lu-zero:
It is an actual bugfix that prevents you from generating invalid code that would silently compile and then break horribly at runtime, I was waiting for https://github.com/eqrion/cbindgen/pull/545 to land in a release and then release `cargo-c` with at least that version of cbindgen since it quite severe.

rustc should also reject that such code in the next stable.

Issue #12, opened 2019-06-11
Add a mean to customize the pkg-config generation


lu-zero:
- [ ] Draft `Cargo.toml` metadata entries
- [ ] Implement the support for it
- [ ] Write a testcase

sdroege:
What I would need here is:
- setting a different name
- setting a different version

For that something like the following would work for me

```toml
[package.metadata.capi]
pkg_config_name = "libebur128"
pkg_config_version = "1.2.4"
```
Or alternatively in a `[package.metadata.capi.pkg_config]` sub-group.

I could implement those but I don't know what your bigger plans with customization here are.

lu-zero:
I'm a bit wary of changing the version, changing the name using `package.metadata.capi.pkg_config.name` sounds fine.

lu-zero:
Ideally I'd start with the `..pkg_config.` keys to allow changing
- name
- description

sdroege:
> I'm a bit wary of changing the version

Why? That's the main thing that I need to change :) My crate just provides an API compatible with a C library but I don't want to follow their versioning scheme.

Implementation plan sounds fine though, I can work on that in the next days.

lu-zero:
I'd keep the pkg config version always identical to the library version if possible, overriding the library version might require a patch in cargo to express it, or to rewrite the target completely though. Let's think more on what can go wrong in one way or another. 

Maybe I'm just too wary and an override here is the easiest path.

sdroege:
Well, overriding it for the pkg-config file only doesn't need a lot of changes and only has an effect if someone actually does that. I don't see the problem :)

lu-zero:
Ok, I'm convinced.

PR #106, opened 2020-08-24
Add support for overriding pkg-config name/description/version

sdroege:
Fixes https://github.com/lu-zero/cargo-c/issues/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,18 @@ subdirectory = true
 generation = true
 ```
 
+### `pkg-config` File Generation
+
+```toml
+[package.metadata.capi.pkg_config]
+# Used as the package name in the pkg-config file and defaults to the crate name.
+name = "libfoo"
+# Used as the package description in the pkg-config file and defaults to the crate description.
+description = "some description"
+# Used as the package version in the pkg-config file and defaults to the crate version.
+version = "1.2.3"
+```
+
 ## Users
 
 - [gcode-rs](https://github.com/Michael-F-Bryan/gcode-rs)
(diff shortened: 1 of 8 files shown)


lu-zero:
I'll test sled, nnnoiseless and rav1e in a bit (or tomorrow) and then I'll land it.

sdroege:
Ok, thanks :) I tested with nnnoiseless and it looked good to me (no change), but please double-check.

For that one I also wonder if it wouldn't make sense to rename the pkg-config file and other things so it is really a drop-in replacement for rnnoise. Currently it provides the same API/ABI, but that's about it. @jneem what do you think?

lu-zero:
You want to have different library name and different pkg-config name so you can install both at the same place. Ideally we could/should add the overrides to the command line as well ^^;

lu-zero:
(once https://github.com/clap-rs/clap/pull/1878 is complete it would be incredibly simple to override from cli)

sdroege:
> You want to have different library name and different pkg-config name so you can install both at the same place. Ideally we could/should add the overrides to the command line as well ^^;

But if the pkg-config name is different, how would existing software pick up nnnoiseless instead of rnnoise without changes in their build system?

lu-zero:
The open question is who is picking. I'd give the option to pass the override from the cli to allow all the options (replace completely or allow side-by-side install).

The default should be non-colliding IMHO.

jneem:
My inclination is to require opt-in for replacing rnnoise with nnnoiseless, especially if it's just a matter of changing `pkg-config rnnoise` to `pkg-config nnnoiseless`.

PR #107, opened 2020-08-24
Run grcov on ubuntu-nightly only

Luni-4:
- Reduce the CI time a bit
- Coverage cross-platform code only


lu-zero:
Fine for me :)

PR #109, opened 2020-09-04
Add more platforms support

tguillem:
Only tested with static build.

### DIFF
--- a/src/build_targets.rs
+++ b/src/build_targets.rs
@@ -32,7 +32,7 @@ impl BuildTargets {
         let env = &target.env;
 
         let (shared_lib, static_lib, impl_lib, def) = match (os.as_str(), env.as_str()) {
-            ("linux", _) | ("freebsd", _) | ("dragonfly", _) | ("netbsd", _) => {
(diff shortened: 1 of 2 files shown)


vibhoothiiaanand:
Another way would be trying to use linux-musl when compiling
It works from adb shell the direct binary^^,
For rav1e we are trying to do that way

lu-zero:
For shared libraries we should not put the version in the soname iirc, somebody can confirm it?

@vibhoothiiaanand it is about building for android, not running on android :)

src/build_targets.rs
```diff
@@ -32,12 +32,12 @@ impl BuildTargets {
         let env = &target.env;
 
         let (shared_lib, static_lib, impl_lib, def) = match (os.as_str(), env.as_str()) {
-            ("linux", _) | ("freebsd", _) | ("dragonfly", _) | ("netbsd", _) => {
+            ("linux", _) | ("freebsd", _) | ("dragonfly", _) | ("netbsd", _) | ("android", _) => {
```
lu-zero, line 35: I confirm you need a different link line, I'll add it now.

lu-zero:
Merged from the other pull requests.

PR #110, opened 2020-09-04
Fix CI binaries

tguillem:
- Add missing cargo-capi on linux/macos binary archive
- Remove target/release path on windows binary archive, to use the same file structure than other OSes.


lu-zero (approved):
It looks fine to me.

PR #111, opened 2020-09-07
Add iOS support

tguillem:
I only tested it with static builds.


lu-zero (approved):

lu-zero:
It looks good to me. Thank you!

PR #112, opened 2020-09-07
CI: Use vendored OpenSSL for macOS release builds

ePirat:
This prevents the release builds to depend on just the right homebrew installed version of OpenSSL on the system running the binaries.


lu-zero (approved):
Looks fine to me

PR #113, opened 2020-09-13
Add minimal example project (including CI)

mgeier:
I think it might be interesting to have an example project (or multiple?) as part of the docs.

On top of being just an example, this could also show how to use `cargo-c` via Github Actions.

The main reason I've made this PR is that there is an error when using MSVC:

```
    Building pkg-config file
    Building .def file using dumpbin
Error: CliError { error: Some(The system cannot find the file specified. (os error 2)), exit_code: 101 }
```

Does this mean that `dumpbin` is not available?
Or is the error talking about another file?

For now, this PR is a WIP because ~of the mentioned error, but also because~ I think it would make sense to add an example C program that uses the provided API.

Of course the example can be arbitrarily expanded, but for now I'm just trying to get a minimal example that compiles successfully on the targeted platforms.

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -35,12 +35,12 @@ jobs:
         run: |
           cargo install --path .
 
-      - name: Set MSVC/dumpbin path
+      - name: Set MSVC binaries path
         if: matrix.toolchain-suffix == '-msvc'
         run: |
-          $MsvcPath = vswhere -latest -products * -find "VC\Tools\MSVC\*\bin\Hostx64\x64" |
-                      Select-Object -Last 1
-          echo "::add-path::$MsvcPath"
+          $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
+          $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
+          echo "$BinPath" >> $GITHUB_PATH
 
       - name: Build example project
         working-directory: example-project
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -35,13 +35,25 @@ jobs:
         run: |
           cargo install --path .
 
+      - name: Set MSVC/dumpbin path
+        if: matrix.toolchain-suffix == '-msvc'
+        run: |
+          $MsvcPath = vswhere -latest -products * -find "VC\Tools\MSVC\*\bin\Hostx64\x64" |
+                      Select-Object -Last 1
+          echo "::add-path::$MsvcPath"
+
       - name: Build example project
         working-directory: example-project
         run: |
           cargo cbuild --verbose --release
 
-      - name: Install example project
+      - name: Install into temporary location
         working-directory: example-project
         run: |
-          cargo cinstall --verbose --release --prefix=/usr/local --destdir=temp
+          cargo cinstall --verbose --release --destdir=temp
+
+      - name: Copy installed files to /usr/local
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        working-directory: example-project
+        run: |
           sudo cp -r temp/usr/local/* /usr/local/
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -40,7 +40,7 @@ jobs:
         run: |
           $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
           $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
-          echo "$BinPath" >> $GITHUB_PATH
+          echo "$BinPath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Build example project
         working-directory: example-project
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -39,6 +39,7 @@ jobs:
         if: matrix.toolchain-suffix == '-msvc'
         run: |
           $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
+          $env:PATH = "$env:PATH;${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer"
           $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
           echo "$BinPath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -1,0 +1,47 @@
+name: Build example project
+
+on: [push, pull_request]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+
+  example-project:
+
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+          - os: windows-latest
+            toolchain-suffix: -gnu
+          - os: windows-latest
+            toolchain-suffix: -msvc
+          - os: ubuntu-latest
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable${{ matrix.toolchain-suffix }}
+          override: true
+
+      - name: Install cargo-c applet
+        run: |
+          cargo install --path .
+
+      - name: Build example project
+        working-directory: example-project
+        run: |
+          cargo cbuild --verbose --release
+
+      - name: Install example project
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --prefix=/usr/local --destdir=temp
+          sudo cp -r temp/usr/local/* /usr/local/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+/example-project/target/
--- a/example-project/Cargo.toml
+++ b/example-project/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "example-project"
+version = "0.1.0"
+edition = "2018"
+
+[features]
+default = ["capi"]
+capi = ["libc"]
+
+[dependencies]
+libc = { version = "0.2", optional = true }
--- a/example-project/Cargo.toml
+++ b/example-project/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2018"
 
 [features]
-default = ["capi"]
+default = []
 capi = ["libc"]
 
 [dependencies]
--- a/example-project/cbindgen.toml
+++ b/example-project/cbindgen.toml
@@ -1,0 +1,4 @@
+include_guard = "EXAMPLE_PROJECT_H"
+include_version = true
+language = "C"
+cpp_compat = true
--- a/example-project/src/capi.rs
+++ b/example-project/src/capi.rs
@@ -1,0 +1,11 @@
+use std::ffi::CStr;
+
+use libc::c_char;
+
+#[no_mangle]
+pub unsafe extern "C" fn example_project_hello(name: *const c_char) {
+    match CStr::from_ptr(name).to_str() {
+        Ok(name) => crate::hello(name),
+        Err(e) => eprintln!("Error: invalid name: {}", e),
+    }
+}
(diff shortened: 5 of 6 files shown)


lu-zero:
Sounds a good idea, also if you are willing to send a PR to add more context to the execution error would be great :)

mgeier:
> send a PR to add more context to the execution error

I was trying this but I don't really understand how the different parts interact here.

I did find out that the `CliError` comes from the `cargo` crate: https://docs.rs/cargo/0.47.0/cargo/util/errors/struct.CliError.html

This `CliError` contains an `anyhow` error. I can add "context" to this `anyhow` error, but it will still not be displayed on the terminal.

The `cbuild()` function returns an `anyhow::Result`, which somehow (but I don't know where and how) this is wrapped into a `CliError` which is then apparently displayed (probably using its `Debug` representation).

How can I get the actual `anyhow` error to be displayed?

mgeier:
It looks like `dumpbin` was indeed not found.

I've managed to add its path in 193868d192c511e03b4d0a36ff3ef29fd2c2926f.

If somebody knows a better way to do this, please let me know!

lu-zero:
@Luni-4 do you have opinions?

lu-zero:
> How can I get the actual `anyhow` error to be displayed?

I should see how cargo is doing and repeat it, CliError is from cargo itself.

.github/workflows/example-project.yml
```diff
@@ -35,13 +35,25 @@ jobs:
         run: |
           cargo install --path .
 
+      - name: Set MSVC/dumpbin path
```
Luni-4, line 38: I would replace the entire block with

```
    - name: Set MSVC x86_64 linker path
      if: matrix.toolchain-suffix == '-msvc'
      run: |
        $LinkGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
        $env:PATH = "$env:PATH;${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer"
        $LinkPath = vswhere -latest -products * -find "$LinkGlob" | Select-Object -Last 1
        echo "::add-path::$LinkPath"
```

.github/workflows/example-project.yml
```diff
@@ -35,13 +35,25 @@ jobs:
...
           cargo cbuild --verbose --release
 
-      - name: Install example project
+      - name: Install into temporary location
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --destdir=temp
+
+      - name: Copy installed files to /usr/local
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
```
Luni-4, line 56: I would replace it with: `if: matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest'`

Luni-4 (changes requested):
Thanks a lot for your work! Just small nits :)

.github/workflows/example-project.yml
mgeier: Sure, I can do that.

Just out of curiosity: why is that better?

Is that documented somewhere?

.github/workflows/example-project.yml
mgeier: OK, I can do that, but why is that better?

What if `macos` or `ubuntu` versions are added/changed later?

.github/workflows/example-project.yml
Luni-4: The first line contains the path where `msvc` binaries are sought. I decided to assign the path to a new variable because that makes easier to change it eventually.
The second line adds the `vswhere` path to the `PATH` environment variable. 
The third and fourth lines are identical to the ones you have written.

.github/workflows/example-project.yml
Luni-4: Because we want to run the tests only on the last `ubuntu` and `macos` versions, not even the other ones.

.github/workflows/example-project.yml
mgeier: > I decided to assign the path to a new variable because that makes easier to change it eventually.

OK, that makes sense.
This way, the `vswhere` command can be written without a line break, which is nicer.

However, the variable name (containing "link") is confusing, isn't it?
AFAIU, this isn't really specific to "linking", but generally about a bunch of binaries (one of which I guess might be the linker?).

More specifically this is *not* a path used by the linker for looking up libraries, right?

> The second line adds the `vswhere` path to the `PATH` environment variable.

Is this necessary?
It seems to work fine without it.

> The third and fourth lines are identical to the ones you have written.

Except the variable name I used is more descriptive.

Luni-4: > However, the variable name (containing "link") is confusing, isn't it?
> AFAIU, this isn't really specific to "linking", but generally about a bunch of binaries (one of which I guess might be the linker?).
> 
> More specifically this is _not_ a path used by the linker for looking up libraries, right?

Yep, exactly. Perhaps we may call it `VSBinariesGlob`, since it refers to the path containing `Visual Studio` binaries. What do you think? 

> Is this necessary?
> It seems to work fine without it.

Let's say it's an additional protection in case `GitHub` decides to remove the `vswhere` path from the `PATH` environment variable. Feel free to remove it if it works the same :)

> 
> Except the variable name I used is more descriptive.

Something line `VSBinariesPath`?

mgeier: I've added 1ab0e6fce226daf6890cd7a841362fb0d9c65636.

I chose the shorter name `$BinGlob`, because otherwise the line would have become overly long.

I've not added the unnecessary `PATH` setting; if it becomes necessary at some point in the future, we can add it then.

I've also changed to the new-style path-setting mechanism, see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/.

mgeier:
I've removed `capi` from the default features in 955a1b42a9094931fae8a335ad0082ef4898c7c9 because it doesn't seem to be necessary.

Is this how it is supposed to be used?

.github/workflows/example-project.yml
Luni-4: Perfect! I agree with your changes, thanks a lot! :)

.github/workflows/example-project.yml
Luni-4: What about this one instead?

.github/workflows/example-project.yml
Luni-4: @mgeier 

It seems the CI fails to build on Windows, perhaps you should remove the `"` the `$BinGlob` variable on line 42

mgeier: It looks like setting the path with Powershell is ridiculously complicated, see https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files.

I've updated it in: 9ece0d32476e6823a4c7e5183b473d6cee4b5cce

> perhaps you should remove the `"` from the `$BinGlob` variable on line 42

I assumed this is necessary in case the path contains spaces?

But I'm not a PowerShell nor a Windows user at all, so I have no idea!

Luni-4: You can add `shell: bash` if you fell more comfortable https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

Luni-4:
> I've removed `capi` from the default features in [955a1b4](https://github.com/lu-zero/cargo-c/commit/955a1b42a9094931fae8a335ad0082ef4898c7c9) because it doesn't seem to be necessary.
> 
> Is this how it is supposed to be used?

Yep, it is not necessary, you can remove it for now

.github/workflows/example-project.yml
mgeier: I still don't quite understand why you want that change.

My version works always, your version breaks when other OS versions are selected.

And even if we are totally sure that we will never ever change the OS versions, I still don't see any disadvantage in my version.

More importantly, this is supposed to become something like a template for other people, who probably want to use different OS versions.

.github/workflows/example-project.yml
mgeier: Cool, I didn't know that!

I personally would indeed feel more comfortable with `bash`, but I think it would be good to use the "default" way of doing things in this example config.
I guess Windows users are more comfortable with PowerShell?
And it's the default, anyway.

What do you think?

.github/workflows/example-project.yml
Luni-4: If you want to test this CI script on other operating systems, practically adding more configurations in the matrix, you're perfectly right, but since there are only the `latest` versions for now in the matrix, I thought we could have done a simple comparisons instead of handling strings using the `startsWith` function. 

But now you told me you are interested in adding (in a future) further configurations, so I'm fine with your change :)

.github/workflows/example-project.yml
Luni-4: Yep, it's a valid reason, so let's use the `Powershell`

mgeier:
Any ideas what the Windows/MSVC CI failure is about?

```
Run cargo cinstall --verbose --release --destdir=temp
       Fresh libc v0.2.79
       Fresh example-project v0.1.0 (D:\a\cargo-c\cargo-c\example-project)
    Finished release [optimized] target(s) in 0.33s
    Building pkg-config file
  Installing pkg-config file
  Installing header file
  Installing static library
  Installing shared library
Error: CliError { error: Some(Cannot copy D:\a\cargo-c\cargo-c\example-project\target\.\release\example_project.dll.lib to temp\usr\local\lib\example_project.dll.lib.

Caused by:
    The system cannot find the file specified. (os error 2)), exit_code: 101 }

Luni-4:
> Any ideas what the Windows/MSVC CI failure is about?
> 
> ```
> Run cargo cinstall --verbose --release --destdir=temp
>        Fresh libc v0.2.79
>        Fresh example-project v0.1.0 (D:\a\cargo-c\cargo-c\example-project)
>     Finished release [optimized] target(s) in 0.33s
>     Building pkg-config file
>   Installing pkg-config file
>   Installing header file
>   Installing static library
>   Installing shared library
> Error: CliError { error: Some(Cannot copy D:\a\cargo-c\cargo-c\example-project\target\.\release\example_project.dll.lib to temp\usr\local\lib\example_project.dll.lib.
> 
> Caused by:
>     The system cannot find the file specified. (os error 2)), exit_code: 101 }
> ```

Adding the `vswhere` [path](https://github.com/Luni-4/cargo-c/commit/57f8a002da033ccffc585cd98ece970276462a31) fixed the [problem](https://github.com/Luni-4/cargo-c/runs/1244544716)

mgeier:
> Adding the `vswhere` [path](https://github.com/Luni-4/cargo-c/commit/57f8a002da033ccffc585cd98ece970276462a31) fixed the [problem](https://github.com/Luni-4/cargo-c/runs/1244544716)

Hmmm, that's strange ... because adding it (e8e9baea5b5754e2252a8c326cc3f0ec9afe6a32) doesn't fix it here (https://github.com/lu-zero/cargo-c/pull/113/checks?check_run_id=1246911997)!

I think the `PATH` thing is unrelated, there is something else going on ...

Luni-4:
> > Adding the `vswhere` [path](https://github.com/Luni-4/cargo-c/commit/57f8a002da033ccffc585cd98ece970276462a31) fixed the [problem](https://github.com/Luni-4/cargo-c/runs/1244544716)
> 
> Hmmm, that's strange ... because adding it ([e8e9bae](https://github.com/lu-zero/cargo-c/commit/e8e9baea5b5754e2252a8c326cc3f0ec9afe6a32)) doesn't fix it here (https://github.com/lu-zero/cargo-c/pull/113/checks?check_run_id=1246911997)!
> 
> I think the `PATH` thing is unrelated, there is something else going on ...

Hmm, could you try to squash all commits and force-push the PR? Because my branch is exactly the same

mgeier:
> Because my branch is exactly the same as yours

I know! That's so strange!

> could you try to squash all commits and force-push the PR?

Without squashing, I got a success and a failure for the exact same commit (e8e9baea5b5754e2252a8c326cc3f0ec9afe6a32):

* failure in this repo: https://github.com/lu-zero/cargo-c/pull/113/checks?check_run_id=1246911997
* success in my fork: https://github.com/mgeier/cargo-c/runs/1246911719?check_suite_focus=true

I don't want to destroy the diffs and discussions in this PR. And I wanted to re-base it anyway, so I've made a new branch and a new PR: #122.
Sadly, CI fails:

* failure in this repo: https://github.com/lu-zero/cargo-c/pull/122/checks?check_run_id=1247050821
* failure in my fork: https://github.com/mgeier/cargo-c/runs/1246992114?check_suite_focus=true

I'm closing this PR, to be continued at #122.

PR #116, opened 2020-10-07
Upgrade cbindgen dependency to 0.15

mgeier:
This enables the automatic conversion from  `Box<T>` (and `Option<Box<T>>`) to `*T` , see https://github.com/eqrion/cbindgen/issues/474.

Issue #98, opened 2020-07-31
Problem with cargo-c and alpine 3.12.0 and rust 1.45.1


wader:
Hello!

Im trying to use cargo-c 0.6.10 with alpine 3.12.0 and latest stable rust (1.45.1) but it seems like the cargo-c binaries segfault for some reason. Maybe i'm doing something wrong?

Reproduction steps:
```Dockerfile
FROM alpine:3.12.0

RUN apk add --no-cache build-base curl rustup openssl-dev
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -- /dev/stdin -y
RUN \
    source $HOME/.cargo/env && \
    cargo install cargo-c    
RUN \
    source $HOME/.cargo/env && \
    cargo-cinstall
```
```sh
$ docker build .
...
   Compiling git2-curl v0.14.0
   Compiling cargo v0.45.1
   Compiling cargo-c v0.6.10
    Finished release [optimized] target(s) in 14m 42s
  Installing /root/.cargo/bin/cargo-capi
  Installing /root/.cargo/bin/cargo-cbuild
  Installing /root/.cargo/bin/cargo-cinstall
   Installed package `cargo-c v0.6.10` (executables `cargo-capi`, `cargo-cbuild`, `cargo-cinstall`)
Removing intermediate container 81753412e8c6
 ---> 0921dec8b843
Step 5/5 : RUN     source $HOME/.cargo/env &&     cargo-cinstall
 ---> Running in 4eca87a7a9f8
/bin/sh: cargo-cinstall: not found
The command '/bin/sh -c source $HOME/.cargo/env &&     cargo-cinstall' returned a non-zero code: 127
```
I did some poke around in the container:
```sh
$ cd /root/.cargo/bin/
$ ls -l cargo-cinstall
-rwxr-xr-x    1 root     root      20053992 Jul 31 19:03 cargo-cinstall
$ ./cargo-cinstall
bash: ./cargo-cinstall: No such file or directory
$ file cargo-cinstall
cargo-cinstall: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld64.so.1, with debug_info, not stripped
$ ldd cargo-cinstall
	/lib/ld64.so.1 (0x7f1cdda5c000)
	libssl.so.1.1 => /lib/libssl.so.1.1 (0x7f1cdd9db000)
[...]

PR #122, opened 2020-10-13
Add minimal example project (including CI) - take 2

mgeier:
This is an extension of #113.

I've re-based this PR in order to take advantage of #116 (`Box<T>` gets converted to `*T`).

I've also extended the Rust example code a bit, using an opaque pointer in the C API. I think this is a typical use case, isn't it?
I've also added a test program and a corresponding `Makefile`. I also added a `README.md`.

I think it is good to test the actual linking and usage of the generated library in CI, which is now done for Linux and macOS.

I have no idea how such a thing is typically done in Windows, but I guess this can be added later by somebody else.

I guess an actual C unit test framework could be used to test this, but for now I chose the simpler option of a small hand-written test program.

For now, only a simple `Makefile` is shown as an example for a build system, but at a later point other build systems (like e.g. CMake) could be added.

I've added a CI step that prints out the `pkg-config` results, any ideas how to more properly test this?

The auto-generated header file is installed to `/usr/local/include/example_project/example_project.h`. Is there a way to avoid creating a separate sub-directory and install it directly to `/usr/local/include/example_project.h`?

For now I've used this:

```c
#include <example_project/example_project.h>
```

... but I think it would be more typical (at least *for small projects*) to do this:

```c
#include <example_project.h>
```

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -12,12 +12,12 @@ jobs:
     strategy:
       matrix:
         include:
+          - os: ubuntu-latest
           - os: macos-latest
           - os: windows-latest
             toolchain-suffix: -gnu
           - os: windows-latest
             toolchain-suffix: -msvc
-          - os: ubuntu-latest
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -68,8 +68,9 @@ jobs:
       - name: Test pkg-config
         if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
         run: |
-          pkg-config --cflags example_project
-          pkg-config --libs example_project
+          set -x
+          test "$(pkg-config --cflags example_project)" = "-I/usr/local/include/example_project"
+          test "$(pkg-config --libs example_project)" = "-L/usr/local/lib -lexample_project"
 
       - name: Update dynamic linker cache
         if: startsWith(matrix.os, 'ubuntu')
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -1,0 +1,83 @@
+name: Build example project
+
+on: [push, pull_request]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+
+  example-project:
+
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+          - os: windows-latest
+            toolchain-suffix: -gnu
+          - os: windows-latest
+            toolchain-suffix: -msvc
+          - os: ubuntu-latest
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Clone Git repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable${{ matrix.toolchain-suffix }}
+          override: true
+
+      - name: Install cargo-c applet
+        run: |
+          cargo install --path .
+
+      - name: Set MSVC binaries path
+        if: matrix.toolchain-suffix == '-msvc'
+        run: |
+          $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
+          $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
+          echo "$BinPath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      - name: Test example project
+        working-directory: example-project
+        run: |
+          cargo test --verbose
+
+      - name: Build C API for example project
+        working-directory: example-project
+        run: |
+          cargo cbuild --verbose --release
+
+      - name: Install into temporary location
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --destdir=temp
+
+      - name: Copy installed files to /usr/local
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        working-directory: example-project
+        run: |
+          sudo cp -r temp/usr/local/* /usr/local/
+
+      - name: Test pkg-config
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        run: |
+          pkg-config --cflags example_project
+          pkg-config --libs example_project
+
+      - name: Update dynamic linker cache
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          sudo ldconfig
+
+      - name: Test usage from C (using Makefile)
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        working-directory: example-project/usage-from-c
+        run: |
+          make
--- a/example-project/cbindgen.toml
+++ b/example-project/cbindgen.toml
@@ -1,0 +1,7 @@
+include_guard = "EXAMPLE_PROJECT_H"
+include_version = true
+language = "C"
+cpp_compat = true
+
+[export.rename]
+"OddCounter" = "ExampleProjectOddCounter"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
 **/*.rs.bk
 Cargo.lock
+/example-project/target/
+/example-project/makefile-usage/run_tests
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 **/*.rs.bk
 Cargo.lock
 /example-project/target/
-/example-project/makefile-usage/run_tests
+/example-project/usage-from-c/run_tests
--- a/example-project/Cargo.toml
+++ b/example-project/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "example-project"
+version = "0.1.0"
+edition = "2018"
+
+[features]
+default = []
+capi = ["libc"]
+
+[dependencies]
+libc = { version = "0.2", optional = true }
--- a/example-project/README.md
+++ b/example-project/README.md
@@ -1,0 +1,19 @@
+Example project using `cargo-c`
+===============================
+
+For detailed usage instructions, have a look at the
+[Github workflow configuration](../.github/workflows/example-project.yml).
+
+Note that `cargo install --path .` is used to install `cargo-c`
+from the locally cloned Git repository.
+If you want to install the latest release from
+[crates.io](https://crates.io/crates/cargo-c),
+you should use this instead:
+
+    cargo install cargo-c
+
+Running `cargo cbuild --release` will create the C header file
+`example_project.h` in the `target/release` directory.
+This file will contain the comments from the file [`capi.rs`](src/capi.rs).
+
+Run `cargo doc --open` to view the documentation of the Rust code.
--- a/example-project/src/capi.rs
+++ b/example-project/src/capi.rs
@@ -1,0 +1,31 @@
+use crate::OddCounter;
+
+// NB: The documentation comments from this file will be available
+//     in the auto-generated header file example_project.h
+
+/// Create new counter object given a start value.
+///
+/// On error (if an even start value is used), NULL is returned.
+/// The returned object must be eventually discarded with example_project_oddcounter_free().
+#[no_mangle]
+pub extern "C" fn example_project_oddcounter_new(start: u32) -> Option<Box<OddCounter>> {
+    OddCounter::new(start).ok().map(Box::new)
+}
+
+/// Discard a counter object.
+///
+/// Passing NULL is allowed.
+#[no_mangle]
+pub extern "C" fn example_project_oddcounter_free(_: Option<Box<OddCounter>>) {}
+
+/// Increment a counter object.
+#[no_mangle]
+pub extern "C" fn example_project_oddcounter_increment(counter: &mut OddCounter) {
+    counter.increment()
+}
+
+/// Obtain the current value of a counter object.
+#[no_mangle]
+pub extern "C" fn example_project_oddcounter_get_current(counter: &OddCounter) -> u32 {
+    counter.current()
+}
--- a/example-project/src/lib.rs
+++ b/example-project/src/lib.rs
@@ -1,0 +1,68 @@
+/*!
+Example library for [cargo-c].
+
+[cargo-c]: https://crates.io/crates/cargo-c
+*/
+
+#![warn(rust_2018_idioms)]
+#![deny(missing_docs)]
+
+#[cfg(cargo_c)]
+mod capi;
+
+/// A counter for odd numbers.
+///
+/// Note that this `struct` does *not* use `#[repr(C)]`.
+/// It can therefore contain arbitrary Rust types.
+/// In the C API, it will be available as an *opaque pointer*.
+#[derive(Debug)]
+pub struct OddCounter {
+    number: u32,
+}
+
+impl OddCounter {
+    /// Create a new counter, given an odd number to start.
+    pub fn new(start: u32) -> Result<OddCounter, OddCounterError> {
+        if start % 2 == 0 {
+            Err(OddCounterError::Even)
+        } else {
+            Ok(OddCounter { number: start })
+        }
+    }
+
+    /// Increment by 2.
+    pub fn increment(&mut self) {
+        self.number += 2;
+    }
+
+    /// Obtain the current (odd) number.
+    pub fn current(&self) -> u32 {
+        self.number
+    }
+}
+
+/// Error type for [OddCounter::new].
+///
+/// In a "real" library, there would probably be more error variants.
+#[derive(Debug)]
+pub enum OddCounterError {
+    /// An even number was specified as `start` value.
+    Even,
+}
+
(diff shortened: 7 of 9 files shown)


Luni-4:
@mgeier 

Thanks for your changes! :)

Could you please rebase?

To fix this CI problem instead: 
```
Error: CliError { error: Some(Cannot copy D:\a\cargo-c\cargo-c\example-project\target\.\release\example_project.dll.lib to temp\usr\local\lib\example_project.dll.lib.

Caused by:
    The system cannot find the file specified. (os error 2)), exit_code: 101 }
Error: Process completed with exit code 1.
```

I guess we need to create the `temp` folder used by `cinstall` manually.

mgeier:
> Could you please rebase?

Done.

> I guess we need to create the `temp` folder used by `cinstall` manually.

Shouldn't `cargo cinstall` take care of this automatically?

Luni-4:
> I've also extended the Rust example code a bit, using an opaque pointer in the C API. I think this is a typical use case, isn't it?

Yep, it's a good solution

> I have no idea how such a thing is typically done in Windows, but I guess this can be added later by somebody else.

Yes, we can extend it in another PR
> 
> For now, only a simple `Makefile` is shown as an example for a build system, but at a later point other build systems (like e.g. CMake) could be added.

i think a `Makefile` is enough, it's not necessary to add an additional build systems for a single example file 

> I've added a CI step that prints out the `pkg-config` results, any ideas how to more properly test this?

For now it's fine, we could eventually add a test to compare the `pkg-config` output, but in another PR

 
> The auto-generated header file is installed to `/usr/local/include/example_project/example_project.h`. Is there a way to avoid creating a separate sub-directory and install it directly to `/usr/local/include/example_project.h`?
> 
> For now I've used this:
> 
> ```c
> #include <example_project/example_project.h>
> ```
> 
> ... but I think it would be more typical (at least _for small projects_) to do this:
> 
> ```c
> #include <example_project.h>
> ```

You can add `pkg-config` headers and libs paths to the `Makefile`.

Something like:
```
CFLAGS=`pkg-config --cflags example_project`
LDFLAGS=`pkg-config --libs example_project`
CC=gcc

all: example_project

example_project: example_project.o
     $(CC) -o $@ $(CFLAGS) $(LDFLAGS) $< 
```

In this way you can use `#include <example_project.h>` in your file.

Luni-4:
> Shouldn't `cargo cinstall` take care of this automatically?

Nope, the default value for `destdir` is this [one](https://github.com/lu-zero/cargo-c/blob/53a4f6b66fe5c8af49c44963dee09115b974708a/src/install.rs#L199) while the directories created by `cinstall` are these [ones](https://github.com/lu-zero/cargo-c/blob/53a4f6b66fe5c8af49c44963dee09115b974708a/src/install.rs#L100-L103)

mgeier:
@Luni-4 Thanks for the review!

In the meantime, I've renamed the directory with the C test: e0552029d7c06c1d594dc613cb70ed7af519b241.
And I added a little more documentation: a86748bc4bc2f6347ed4cc45bf30d01d5788ff60.

> > Shouldn't `cargo cinstall` take care of this automatically?
> 
> Nope, the default value for `destdir` is this [one](https://github.com/lu-zero/cargo-c/blob/53a4f6b66fe5c8af49c44963dee09115b974708a/src/install.rs#L199) while the directories created by `cinstall` are these [ones](https://github.com/lu-zero/cargo-c/blob/53a4f6b66fe5c8af49c44963dee09115b974708a/src/install.rs#L100-L103)

I've decided this is too complicated for me (as a non-Windows-user), so I disabled the installation on Windows for now: 799aacbc3ef10f24ee8ffc9749ac18bbae557b4d.
I hope that somebody else can fix this later.

> > I have no idea how such a thing is typically done in Windows, but I guess this can be added later by somebody else.
> 
> Yes, we can extend it in another PR

Perfect, this goes way beyond my abilities (and I don't really want to learn how to do it).

> > For now, only a simple `Makefile` is shown as an example for a build system, but at a later point other build systems (like e.g. CMake) could be added.
> 
> i think a `Makefile` is enough, it's not necessary to add an additional build system for a single example file

OK, cool.

I thought that maybe CMake could be used, because this might also work on Windows ... but this is beyond the scope of this PR.

> > I've added a CI step that prints out the `pkg-config` results, any ideas how to more properly test this?
> 
> For now it's fine, we could eventually add a test to compare the `pkg-config` output, but in another PR

OK, cool.

> > The auto-generated header file is installed to `/usr/local/include/example_project/example_project.h`. Is there a way to avoid creating a separate sub-directory and install it directly to `/usr/local/include/example_project.h`?
> > For now I've used this:
> > ```c
> > #include <example_project/example_project.h>
> > ```
> > 
> > 
> > ... but I think it would be more typical (at least _for small projects_) to do this:
> > ```c
> > #include <example_project.h>
> > ```
> 
> You can add `pkg-config` headers and libs paths to the `Makefile`.

I know, but if the library is installed to the default location, I don't even need `pkg-config`.

And I don't see any advantage in having the header file in a sub-directory.

The only situation where this makes sense is if there are multiple header files.
Does `cargo-c` ever create multiple header files?

I think the default for single header files should be to *not* create a sub-directory.

Anyway, this can be further discussed and implemented at another time.
From my side, this PR is ready to be merged.
Is there still something missing?

Luni-4:
@mgeier Thanks to you for your work! 

Yes, we can implement the `Windows` stuff later. 

> I thought that maybe CMake could be used, because this might also work on Windows ... but this is beyond the scope of this PR.

Thinking about this more in deep, building the example with a build system could help people to figure out how to use it alongside `cargo-c`, so if you are interested in it, feel free to add that in another PR. I would advise the `meson` build system for its much simpler syntax.

> I know, but if the library is installed to the default location, I don't even need `pkg-config`.
> 
> And I don't see any advantage in having the header file in a sub-directory.
> 
> The only situation where this makes sense is if there are multiple header files.
> Does `cargo-c` ever create multiple header files?
> 
> I think the default for single header files should be to _not_ create a sub-directory.
> 
> Anyway, this can be further discussed and implemented at another time.

Can we use the sub-directory approach for this example for now? As you said, we can always change that behavior later.

> From my side, this PR is ready to be merged.
> Is there still something missing?

After the new sub-directories `Makefile` approach is ready, please squash all commits into a single one and I'm going to review your work :)

mgeier:
> building the example with a build system could help people to figure out how to use it alongside `cargo-c`, so if you are interested in it, feel free to add that in another PR. I would advise the `meson` build system for its much simpler syntax.

I think this is a good idea, but I'm not familiar with `meson` (nor with any other cross-platform build system), so somebody else should tackle this.

> Can we use the sub-directory approach for this example for now?

Sure, currently that's the only way to do it, AFAICT.
For now, the header file will end up in `/usr/local/include/example_project/example_project.h`.

Getting rid of the unnecessary sub-directory is a separate task, I've just created an issue for that: #125.

> After the new sub-directories `Makefile` approach is ready,

I don't understand what you mean by that.
Currently, the `#include` contains the sub-directory, which is IMHO the most reasonable thing to do until #125 is solved.

> please squash all commits into a single one and I'm going to review your work :)

Done.

Luni-4:
> > Can we use the sub-directory approach for this example for now?
> 
> Sure, currently that's the only way to do it, AFAICT.
> For now, the header file will end up in `/usr/local/include/example_project/example_project.h`.
> 
> Getting rid of the unnecessary sub-directory is a separate task, I've just created an issue for that: #98 

Perfect, thanks! 

> I don't understand what you mean by that.
> Currently, the `#include` contains the sub-directory, which is IMHO the most reasonable thing to do until #125 is solved.

Using `pkg-config` in the `Makefile`, so you can use `#include <example_project.h>` instead of `#include <example_project/example_project.h>` in the `C` file

mgeier:
> Using `pkg-config` in the `Makefile`, so you can use `#include <example_project.h>` instead of `#include <example_project/example_project.h>` in the `C` file

I don't think that's the best way to do it in this instructional example.

Since the library is installed into the default location, no `pkg-config` should be necessary.

If we *would* want to show the use of `pkg-config`, we should install the example library into a non-standard location. But why would we want to do that? Furthermore, we would also have to specify the path to `pkg-config` itself in the `Makefile` (which typically involves a `PKG_CONFIG` variable), which would unnecessarily complicate the situation.

I think we should show the most typical usage in this example, therefore I chose not to involve `pkg-config`.

The switch to `#include <example_project.h>` should only happen once #125 is resolved.

Luni-4:
> > Using `pkg-config` in the `Makefile`, so you can use `#include <example_project.h>` instead of `#include <example_project/example_project.h>` in the `C` file
> 
> I don't think that's the best way to do it in this instructional example.
> 
> Since the library is installed into the default location, no `pkg-config` should be necessary.
> 
> If we _would_ want to show the use of `pkg-config`, we should install the example library into a non-standard location. But why would we want to do that? Furthermore, we would also have to specify the path to `pkg-config` itself in the `Makefile` (which typically involves a `PKG_CONFIG` variable), which would unnecessarily complicate the situation.
> 
> I think we should show the most typical usage in this example, therefore I chose not to involve `pkg-config`.
> 
> The switch to `#include <example_project.h>` should only happen once #125 is resolved.

Seems legit, let's fix #125 first

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+  example-project:
+
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+          - os: windows-latest
+            toolchain-suffix: -gnu
+          - os: windows-latest
+            toolchain-suffix: -msvc
```
Luni-4, line 19: We are not considering `Windows` anymore, so we can remove these lines.

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+
+    steps:
+      - name: Clone Git repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable${{ matrix.toolchain-suffix }}
```
Luni-4, line 32: Leave only `toolchain: stable`

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
+          - os: windows-latest
+            toolchain-suffix: -gnu
+          - os: windows-latest
+            toolchain-suffix: -msvc
+          - os: ubuntu-latest
```
Luni-4, line 20: Move this line above `os: macos-latest`

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Install cargo-c applet
+        run: |
+          cargo install --path .
+
+      - name: Set MSVC binaries path
+        if: matrix.toolchain-suffix == '-msvc'
+        run: |
+          $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
+          $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
+          echo "$BinPath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
Luni-4, line 44: Remove all these lines.

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          cargo test --verbose
+
+      - name: Build C API for example project
+        working-directory: example-project
+        run: |
+          cargo cbuild --verbose --release
+
+      - name: Install into temporary location
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
```
Luni-4, line 57: No more necessary

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --destdir=temp
+
+      - name: Copy installed files to /usr/local
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
```
Luni-4, line 63: No more necessary

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+          pkg-config --cflags example_project
+          pkg-config --libs example_project
+
+      - name: Update dynamic linker cache
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          sudo ldconfig
+
+      - name: Test usage from C (using Makefile)
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
```
Luni-4, line 80: No more necessary

.github/workflows/example-project.yml
```diff
@@ -0,0 +1,83 @@
...
+        working-directory: example-project
+        run: |
+          sudo cp -r temp/usr/local/* /usr/local/
+
+      - name: Test pkg-config
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
+        run: |
+          pkg-config --cflags example_project
+          pkg-config --libs example_project
```
Luni-4, line 72: Since we have decided not to consider `pkg-config` until #125 will be fixed, we can remove these lines too.

example-project/cbindgen.toml
```diff
@@ -0,0 +1,7 @@
+include_guard = "EXAMPLE_PROJECT_H"
+include_version = true
+language = "C"
+cpp_compat = true
+
+[export.rename]
+"OddCounter" = "ExampleProjectOddCounter"
```
Luni-4, line 7: Can we always use `OddCounter`?

Luni-4 (changes requested):
@mgeier 

I addressed the points discussed above in this review

.github/workflows/example-project.yml
mgeier: For me, this is the most important piece of information in this whole PR.

If I had had this information in the first place, I probably wouldn't even have created this PR.

I think it is extremely valuable and non-obvious.

.github/workflows/example-project.yml
mgeier: This test already checks whether the `pkg-config` configuration is found (which is better than nothing), but it doesn't check whether the output is correct.

Instead of removing the partial check, I decided to complete the check in e4784179e316cefcbb19f5c3014e08d0486600b8.

example-project/cbindgen.toml
mgeier: Sure, we could.

But AFAIK it is common for a C API to prefix every public symbol with the project name (to avoid name clashes), so that's what I did here.
Also, I think the `export.rename` feature is really useful, so it's nice to show it here.

But if you don't like it, I can remove it.

Luni-4: Seems a good reason, let's leave it as-is

.github/workflows/example-project.yml
Luni-4: > I thought we were only postponing compiling and running a C program on Windows.

Yep, so let's add those configurations in another PR, in that way we only have `Linux` and `macOS` stuff in this one. 

> But I think there is already some value in running `cargo cbuild` on Windows (both GNU and MSVC), isn't it?
> 
> In fact, the whole reason why I started working on this topic (#113) is a build failure on MSVC, which has been fixed by specifying the MSVC binaries path, which was a really helpful information for my own project, and I think it would be helpful for others as well.

We are not dealing with `Windows` stuff in this PR, so I propose to add the `cargo cbuild` check in another PR

.github/workflows/example-project.yml
Luni-4: Same reason used above for this one

.github/workflows/example-project.yml
Luni-4: Ok, sounds good to me. Can `set -x` be removed?

.github/workflows/example-project.yml
mgeier: For me, Windows is the main reason why I even created this PR.

What is the advantage of removing the Windows stuff from this PR?

To me, this just seems like unnecessary work.

.github/workflows/example-project.yml
mgeier: Sure, it *could* be removed. But then we would miss valuable information in the error case.

Luni-4: Ok, no problem then

.github/workflows/example-project.yml
Luni-4: Since most of the `Windows` things won't be dealt dealt in this PR (due to the previous messages above), I'd rather handle **all** `Windows` stuff in a completely different PR, even the `cargo cbuild` check. I'm not saying to discard all your work, just move and extend it in another PR.

In that way, it will be much easier to address problems since they refer to different concerns.

mgeier: > move and extend it in another PR.

I will not be able to extend it, so it won't be possible to do "**all** `Windows` stuff in a completely different PR" anyway (unless my contribution is discarded).
So there is no reason to split this PR into two parts.

I think the Windows stuff that I've added here is useful on its own, that fact that it is not "complete" doesn't matter to me.

And if I *would* split this PR into two, the second one would be "incomplete", and this whole discussion would start again ...

> In that way, it will be much easier to address problems since they refer to different concerns.

I don't believe that's true.

If you see problems with this PR, I'm happy to try and fix them, it not, let's merge!

Luni-4: Someone else will iterate over this in another PR, fine!

mgeier:
Thanks for merging!
I've created an issue as a reminder for the Windows stuff: #128.

PR #124, opened 2020-10-13
Fix 'index out of bounds' in build.rs

JonasVautherin:
I am getting the following error:

```
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1'
```

Looking at the code, it seems that if `targets.len()` is 1, then the first element should be `targets[0]` and not `targets[1]`.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -407,7 +407,7 @@ pub fn cbuild(
     let targets = args.targets();
     let target = match targets.len() {
         0 => None,
-        1 => Some(targets[1].clone()),
+        1 => Some(targets[0].clone()),
         _ => {
             anyhow::bail!("Multiple targets not supported yet");
         }


lu-zero:
How do you hit this?

lu-zero (approved):
Looks correct, thank you for spotting it.

JonasVautherin:
Trying to build VLCKit, I end up on this error. I don't know which dependency depends on cargo-c, but it gets your latest release and fails. So I'm looking forward to the next patch release of cargo-c, because I don't know how to fix that otherwise :sweat_smile:.

For the record, I'm debugging my build here: https://code.videolan.org/videolan/VLCKit/-/issues/437

lu-zero:
I'll bump it in few minutes. Thanks for catching it

JonasVautherin:
Thanks a lot, that fixes my VLCKit build!

PR #131, opened 2020-10-19
Don't build shared libraries for musl targets

### LINKED ISSUE #85: Automatically disable cdylib on musl targets
musl targets appear to default to static-only builds.

bofh69:
Fixes #85


lu-zero (approved):
Looks good, thank you!

Issue #135, opened 2020-11-20
Modules/plugin supports


gdesmott:
`gst-plugins-rs`, the GStreamer Rust plugins, are currently [not statically linkable](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/133) with a C application. One of the issue is the lack of pkg-config file to retrieve its linking flags. [It's been suggested](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/133#note_700866) that `cargo-c` could be a good fit for this as it already has support for pkg-config generation.

Would you be willing to extend `cargo-c` to support such use case?

cc @sdroege @xclaesse

sdroege:
Basically this would mean
 - `build` wouldn't include a soname and stuff
 - `pkg-config` file would be generated
 - no header is generated or installed
 - installation of the generated .so (and pkg-config) should be in a special location, not the standard library path

@lu-zero What do you think?

lu-zero:
Ideally we just need to add a key to ignore the architecture-provided link lines.

Right now I'm updating to the latest cargo, hopefully this one won't make release libraries with llvm-ir in strange .o.

lu-zero:
@gdesmott if you want to give it a try and send a patch I'm quite open to add this feature :)

gdesmott:
Cool! So how would prefer to have this API-wise?

Should we add something like this to disable the so name mangling and header generation?

```toml
[package.metadata.capi]
library = false
```

Or do you prefer to have individual settings?

```toml
[package.metadata.capi.header]
enabled = false

[package.metadata.capi.library]
so_name_versioning = false
```

Regarding the custom install path should we have something like this?

```toml
[package.metadata.capi.library]
install_subdir = "gstreamer-1.0"
```
which would install to `$dest/$libdir/$subdir`.

lu-zero:
> Or do you prefer to have individual settings?
> 
> ```toml
> [package.metadata.capi.header]
> enabled = false
> 
> [package.metadata.capi.library]
> so_name_versioning = false
> ```

I do

> Regarding the custom install path should we have something like this?
> 
> ```toml
> [package.metadata.capi.library]
> install_subdir = "gstreamer-1.0"
> ```
> 
> which would install to `$dest/$libdir/$subdir`.

Sounds good to me. :)

sdroege:
Seems like a good approach to me too

gdesmott:
```toml
[...]

PR #141, opened 2020-11-25
Add options needed to build plugins

gdesmott:
See #135 for the context and API discussion.

### DIFF
--- a/src/pkg_config_gen.rs
+++ b/src/pkg_config_gen.rs
@@ -51,6 +51,13 @@ impl PkgConfig {
             _ => Vec::new(),
         };
 
+        let mut libdir = PathBuf::new();
+        libdir.push("${libdir}");
+        if let Some(subdir) = &capi_config.library.install_subdir {
+            libdir.push(subdir);
+        }
+        let libs = format!("-L{} -l{}", libdir.display(), capi_config.library.name);
+
         PkgConfig {
             name: capi_config.pkg_config.name.clone(),
             description: capi_config.pkg_config.description.clone(),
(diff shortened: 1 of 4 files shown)


lu-zero (approved):
It seems good to me, I'm a little busy today so I'd merge it once I give it a go on sled, rav1e and nnnoiseless. If you already tested it it can be merged now.

sdroege:
Also ebur128 please, it makes use of many of the other configurations that are hopefully not affected by this change :)

gdesmott:
I never build or use any of those crates so I prefer to let you double check I didn't break anything. :)

gdesmott:
Just found out that `Libs` field in the pkg-config file should be updated with the subdirectory as well. Let me fix that before merging. :)

gdesmott:
> Just found out that `Libs` field in the pkg-config file should be updated with the subdirectory as well. Let me fix that before merging. :)

Fixed.

I also stopped settings `Cflags` when we are not generating any headers and added a extra commit preventing the creation of `bin` directory if not needed.

src/pkg_config_gen.rs
```diff
@@ -51,6 +51,23 @@ impl PkgConfig {
             _ => Vec::new(),
         };
 
+        let libdir = if let Some(subdir) = &capi_config.library.install_subdir {
+            format!("{}{}{}", "${libdir}", MAIN_SEPARATOR, subdir)
```
lu-zero, line 55: probably we could avoid using `format!` and just append to the path and convert later.

gdesmott: Indeed that's easier, done.

PR #142, opened 2020-11-25
CI: Add "cargo ctest"

mgeier:
This is marked as "draft" because it doesn't seem to run successfully on Windows.

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -53,6 +53,11 @@ jobs:
         run: |
           cargo cbuild --verbose --release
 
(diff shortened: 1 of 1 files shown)


lu-zero:
The problem with linking could be probably solved in `inline-c`. please open an issue on their side as well.

mgeier:
I don't know anything about linking on Windows, can you please do that?

lu-zero:
I found a workaround, please rebase and we'll land it.

mgeier:
Thanks @lu-zero, now the tests pass!

PR #143, opened 2020-11-25
add settings to define Require(.private) in pc file

### LINKED ISSUE #139: pkg-config: allow to define Requires and Requires.private
The pkg-config generator [has code to generate Requires](https://github.com/lu-zero/cargo-c/blob/master/src/pkg_config_gen.rs#L165) but there is currently no way to use it.

Should we add something like so users can define them manually?

```toml
[package.metadata.capi.pkg_config]
requires = "gstreamer-1.0, gstreamer-base-1.0"
requires_private = "gobject-2.0, glib-2.0 >= 2.56.0, gmodule-2.0"
```

gdesmott:
Fix #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ name = "libfoo"
 description = "some description"
 # Used as the package version in the pkg-config file and defaults to the crate version.
 version = "1.2.3"
+# Used as the Requires field in the pkg-config file, if defined
+requires = "gstreamer-1.0, gstreamer-base-1.0"
+# Used as the Requires.private field in the pkg-config file, if defined
+requires_private = "gobject-2.0, glib-2.0 >= 2.56.0, gmodule-2.0"
 ```
 
 ### Library Generation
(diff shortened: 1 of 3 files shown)


lu-zero (approved):
Looks good to me. 

Using `["gstreamer-1.0", "gstreamer-base-1.0"]` would avoid the split + trim on input, but I guess the overall ergonomy would not improve at all. :)

gdesmott:
Yeah, I figured it was best to make things simpler for users. :)

gdesmott:
@lu-zero : we may need something more flexible for crates which dependencies are either built-in or dynamically linked, see https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/439#note_713208
What do you think?

lu-zero:
In theory we'd need to ask the dependencies what they did in a more robust fashion. Shall we brainstorm a little on zulip?

gdesmott:
I never used Zulip but why not.
@sdroege may be interested discussing this as well.

lu-zero:
https://rust-av.zulipchat.com/#narrow/stream/254255-cargo-c/topic/Extending.20the.20pkg-config.20support for new to it

PR #145, opened 2020-11-27
don't rebuild pc file if cache has not been invalidated

### LINKED ISSUE #104: Rebuilds pkg-config file on `cbuild` and `cinstall`
```
$ cargo cbuild --prefix=(pwd)/install
   Compiling ebur128 v0.1.1 (/home/slomo/Projects/rust/ebur128)
    Finished dev [optimized + debuginfo] target(s) in 2.84s
    Building pkg-config file
$ cargo cinstall --prefix=(pwd)/install
    Finished dev [optimized + debuginfo] target(s) in 2.02s
    Building pkg-config file
  Installing pkg-config file
  Installing header file
  Installing static library
  Installing shared library
```

It seems like it also re-links (or even rebuilds) the library itself. If this can't be reproduced with anything else then I'll debug this / provide a testcase in a bit.

gdesmott:
Fix #104

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -248,7 +248,10 @@ fn fingerprint(build_targets: &BuildTargets) -> anyhow::Result<Option<u64>> {
 
     let mut hasher = DefaultHasher::new();
 
-    let mut paths = vec![&build_targets.include];
+    let mut paths: Vec<&PathBuf> = Vec::new();
+    if let Some(include) = &build_targets.include {
+        paths.push(&include);
+    }
     paths.extend(&build_targets.static_lib);
     paths.extend(&build_targets.shared_lib);
 
--- a/src/build.rs
+++ b/src/build.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::io::{BufRead, BufReader, Write};
+use std::io::{BufRead, BufReader, Read, Write};
 use std::path::PathBuf;
 
 use cargo::core::profiles::Profiles;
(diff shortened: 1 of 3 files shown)


gdesmott:
Actually that's not enough. We need to trigger a rebuild if the cli arguments such as  `--prefix` have changed as well.

sdroege:
> Actually that's not enough. We need to trigger a rebuild if the cli arguments such as `--prefix` have changed as well.

And probably also if some of the configuration from `Cargo.toml` has changed

lu-zero:
Yes, fingerprint() probably needs to be extended a bit to make it work reliably for this use-case ^^;

gdesmott:
> We need to trigger a rebuild if the cli arguments such as `--prefix` have changed as well.

What would the proper way to do that? We have two options here:
- parse the existing `.pc` file and check if its prefix, libdir, etc matches the one passed on the cli.
- Hash the CLI arguments (`InstallPaths`) as well when generating the fingerprint and store it in the target directory.

@lu-zero : which option do you prefer?

lu-zero:
Historically in the previous implementation we did the latter and worked sort of fine.

gdesmott:
Ok, what about this then? It no longer regenerate everything when rebuilding the Rust gst plugins.

> And probably also if some of the configuration from `Cargo.toml` has changed

Looks like any change in `Cargo.toml` triggers a rebuild so we should be good.

lu-zero:
Looks good beside the small problem clippy found. I guess once you address it it is good to go.

gdesmott:
Should be good now.

lu-zero (approved):
Looks good to me :)

PR #146, opened 2020-11-27
example-project: Use #[cfg(cargo_c)] instead of #[cfg(feature = "capi")]

mgeier:
... as suggested in the README.

I'm not sure what's the difference in the end, but I guess it's better to follow the suggestion from the README, right?

### DIFF
--- a/example-project/src/lib.rs
+++ b/example-project/src/lib.rs
@@ -7,7 +7,7 @@ Example library for [cargo-c].
 #![warn(rust_2018_idioms)]
 #![deny(missing_docs)]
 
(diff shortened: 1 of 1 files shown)


lu-zero:
I should update the readme to avoid `cfg(cargo_c)` it is a relic of the first implementation and has a number of shortcomings.

Sorry for the confusion.

mgeier:
OK, thanks for clarifying this!

It makes of course more sense if there is only one way to do this.

lu-zero:
Thank you for reminding me :)

PR #149, opened 2020-12-04
generate -uninstalled .pc files

gdesmott:
This can be used to use the generated library without installing it.
    
Will be required to integrate GStreamer Rust plugins with gst-build as it does not install the artifacts.


lu-zero (commented):
Looks good to me. Thank you!

PR #155, opened 2020-12-22
Use CC to find dumpbin

### LINKED ISSUE #147: dumpbin is not in the PATH by default
I'm currently trying to add cargo-c to a project and I'm compiling with MSVC.

The problem is that `dumpbin` is not in the PATH by default. This has been mentioned in  #113 and added in #71.

I've the following error:
```
    Building pkg-config file
    Building .def file using dumpbin
Error: CliError { error: Some(The system cannot find the file specified. (os error 2)), exit_code: 101 }
```

As a workaround I could add `dumpbin` manually to the PATH, which I find ugly. Or maybe use a `build.rs` to find `dumpbin` and add it to the PATH, which is not ideal too.

Maybe the linker path information obtained from cargo can be used since (at least for VS 2019) `dumpbin.exe` is in the same directory as `link.exe` and `cl.exe` ?

Mikayex:
Fixes #147

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -36,13 +36,6 @@ jobs:
         run: |
           cargo install --path .
 
-      - name: Set MSVC binaries path
-        if: matrix.toolchain-suffix == '-msvc'
-        run: |
-          $BinGlob = "VC\Tools\MSVC\*\bin\Hostx64\x64"
-          $BinPath = vswhere -latest -products * -find "$BinGlob" | Select-Object -Last 1
-          echo "$BinPath" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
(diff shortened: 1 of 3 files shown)


lu-zero (approved):
Looks great, thank you!

mgeier:
Since this is only used on Windows, shouldn't the dependency only be installed on Windows?

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies.

lu-zero:
I plan to use it to add support to build C examples out of box. Hopefully it would not take much time. :)

mgeier:
OK, but then it would only be a "dev" dependency, right?

lu-zero:
No, it will be used for `cargo cbuild --examples` :)

PR #169, opened 2021-02-17
Rework native static libs

### LINKED ISSUE #168: Support no_std targets
Our way to get the static libraries is not accounting for the crate being or not being no_std.

While the correct solution is make so rustc does not try to compile and link the crate to emit the static link line on stderr (see https://github.com/rust-lang/rust/issues/61089), we can probably work around it either by manually checking for the presence of the `no_std` attribute and feed rustc with an empty file with the no_std attribute if it is present.

lu-zero:
It should fix #168 and make its detection a bit more robust.

As bonus it should also pave the way to a better logic to trigger a rebuild if some options are changed across builds.


src/build.rs
```diff
@@ -519,9 +535,66 @@ pub fn compile_options(
...
+                    level: String,
+                }
+
+                if let Ok(msg) = serde_json::from_str::<Message>(s) {
+                    // suppress the native-static-libs messages
+                    if msg.level == "note" {
+                        if msg.message.starts_with("Link against the following native artifacts when linking against this static library") {
+                            Ok(())
+                        } else if msg.message.starts_with("native-static-libs:") {
+                            let (_, link_line) = msg.message.split_at(20);
```
sdroege, line 579: Annoying that github does not allow to add suggestions when reviewing individual commits...

```rust
} else if let Some(link_line) = msg.message.strip_prefix("native-static-libs:") {
```

src/build.rs
```diff
@@ -519,9 +535,66 @@ pub fn compile_options(
...
 
+    compile_opts.build_config.unit_graph = false;
+
     Ok(compile_opts)
 }
 
+#[derive(Default)]
+struct Exec {
+    ran: AtomicBool,
+    link_line: Mutex<String>,
```
sdroege, line 546: Should maybe also be an `Option` until it's actually set?

sdroege (commented):
Looks good to me :) Nicer than before

src/build.rs
lu-zero: thank you :)

src/build.rs
lu-zero: I could have a single field thats Mutex<Option<String>> even if it is a bit more implicit maybe.

src/build.rs
```diff
@@ -266,6 +274,7 @@ impl<'a> FingerPrint<'a> {
             root_output,
             build_targets,
             install_paths,
+            static_libs: "",
```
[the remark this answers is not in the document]
lu-zero: I'd rather avoid one level of indirection.

src/build.rs
lu-zero: I tried and the code gets more convoluted. :(

lu-zero: Also the link line can be empty on purpose. So the code should stay this way.

PR #170, opened 2021-02-17
Patch the target name to the custom name if provided

### LINKED ISSUE #166: Renaming the library is broken
Renaming the library and the pkg-config file seems to be broken. I can only rename the header using `package.metadata.capi.header` in `Cargo.toml`. If I try to use

    [package]
    name = "a"

    [package.metadata.capi.library]
    name = "b"

to rename the library I get

       Compiling a v0.1.0 (/tmp/a)
        Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Building pkg-config file
        Building pkg-config file
        Building header file using cbindgen
    Error: CliError { error: Some(Cannot copy /tmp/a/target/./debug/libb.so to /tmp/a/target/./debug/libb.so.0.1.0.

    Caused by:
        No such file or directory (os error 2)), exit_code: 101 }

It seems the first path is wrong.

Finally `package.metadata.capi.pkg_config` seems to be ignored. The content of the `*.pc` seems to be correct though:

    prefix=/usr/local
    exec_prefix=${prefix}
    libdir=${exec_prefix}/lib
    includedir=${prefix}/include

    Name: b
    Description: 
    Version: 0.1.0
    Libs: -L${libdir} -lb
    Cflags: -I${includedir}/a
    Libs.private: -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc

The complete MWE is here: https://gist.github.com/cschwan/0cb317b682541ed6962dcab65cda1b7a

lu-zero:
Fixes #166.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -94,7 +94,11 @@ fn build_pc_file(
     Ok(())
 }
 
-fn patch_lib_kind_in_target(ws: &mut Workspace, libkinds: &[&str]) -> anyhow::Result<()> {
+fn patch_target(
+    ws: &mut Workspace,
+    libkinds: &[&str],
+    capi_config: &CApiConfig,
+) -> anyhow::Result<()> {
     use cargo::core::compiler::CrateType;
 
     let pkg = ws.current_mut()?;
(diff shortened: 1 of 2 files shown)


sdroege (approved):
Makes sense to me, thanks!

PR #175, opened 2021-03-19
Make it easier to have a crt-static build

lu-zero:


cromefire:
Great! It also solved the problem with caching

lu-zero:
I'll cut a new release soon, thank you for testing!

Issue #184, opened 2021-04-17
Architecture independent install files support


naokiri:
I want support for installing architecture independent files. For instance, I have a usecase to install configuration file under /usr/local/share/myproject/.

Once datadir is supported, it should be easy to add infodir, mandir, localedir etc other common shared files support.
c.f. https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html

lu-zero:
This should be discussed with cargo upstream see https://github.com/rust-lang/cargo/issues/2729 and https://github.com/rust-lang/cargo/issues/2386

naokiri:
I think cargo upstream have consensus that the original cargo doesn't want to install things on system-wide, and puts everything under `~/.cargo`.

C ABI libraries typically uses `/usr/share/` and `/usr/local/share` so I thought it's potentially a scope of this project.

As a owner of the project, it's your call. Feel free to close the issue if this wasn't scope of this project.

lu-zero:
I have no problems in adding `install.rs` support first if you have a specific need for it. 

Can you provide more information on which are your use-cases?

naokiri:
Sorry but I might be missing something. Could you elaborate what `install.rs` support means?

My usage of `/usr/local/share` is for my own project: https://github.com/naokiri/cskk porting an existing library into Rust, and need to add `/usr/local/share/libcskk/rules` directory on install.

lu-zero:
There is a postponed [rfc](https://github.com/rust-lang/rfcs/pull/2376) about it, I might add something along those lines to give enough flexibility. I guess much depends on how early you'd need it.

naokiri:
It is just a personal project that no public user exists yet so it's not in hurry at all. And also I'm hesitant to touch the upstream cargo project since I'm not so used to Rust yet and probably cannot help much.

lu-zero:
I'll try to get something in place that is more or or less along the lines of the rfc. So something called `cinstall.rs` that gets inputs and outputs like `build.rs`.

mdegans:
@naokiri 
I've used `cargo-deb` in conjunction with `cargo-c` to generate a debian package for systemwide install. It's not perfect (solib symlinks need fixing with a postinst script), but it mostly works, If you edit the package manually it can be close to perfect, especially for your own use.

lu-zero:
I'll try to get a prototype ready once I have other internal rework landing :)

lu-zero:
[...]

PR #185, opened 2021-04-17
Datadir support  #184

naokiri:
This change adds support for installing read-only architecture-independent data files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,19 @@ install_subdir = "gstreamer-1.0"
 versioning = false
 ```
 
+### Shared Data Install
+```toml
+[package.metadata.capi.shared.data]
+# Can be used to suppress installing shared data files. Defaults to false.
+enabled = false
+# Relative path from the project root where the shared data file is. 
+# Defaults to "shared/data".  
+origin = "src/data"
+# Relative path from `datadir` where the shared data file is installed. Defaults to the crate name. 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -127,6 +127,19 @@ install_subdir = "gstreamer-1.0"
 versioning = false
 ```
 
+### Shared Data Install
+```toml
+[package.metadata.capi.shared.data]
+# Can be used to suppress installing shared data files. Defaults to false.
+enabled = false
+# Relative path from the project root where the shared data file is. 
+# Defaults to "shared/data".
```
naokiri, line 136: I am very hesitant on using this default. Is there any good well-known directory structure for this kind of file?
At first I wanted to use  `build.rs` outputs so I tried to read `OUT_DIR` ( https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts ) but it wasn't available on running `cargo cinstall`

lu-zero (changes requested):
I'd rather have this discussion with upstream cargo if possible. If nothing happens on their side we can look into it.

README.md
naokiri: This is the origin of the data to be installed, so the destination doesn't matter. 
In this example, `src/data` is copied to $datadir/$install_dir, which is `/usr/local/share/libfoo`

lu-zero:
All the machinery to support this use case is in the master, if you have time we can add the remaining bits this month :)

PR #191, opened 2021-05-05
CI: test example project in Windows/MSYS2

mgeier:
This only seems to work with the GNU toolchain, which I guess makes sense.

I'm not an MSYS2 expert (in fact I've only ever used it on CI servers and I started very recently), so there might be a better way to do this.

It seems that `/usr/local/*` is not searched by default, I don't know why.

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -80,3 +80,29 @@ jobs:
         working-directory: example-project/usage-from-c
         run: |
           make
+
+      - name: Install into temporary location
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --prefix=/mingw64 --destdir=temp
+
+      - name: Install MSYS2
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        uses: msys2/setup-msys2@v2
+        with:
+          release: false
+
+      - name: Copy installed files to /mingw64
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        working-directory: example-project
+        run: |
+          cp -r temp/mingw64/* /mingw64/
+        shell: msys2 {0}
+
+      - name: Test usage from C (using Makefile)
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        working-directory: example-project/usage-from-c
+        run: |
+          make
+        shell: msys2 {0}
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -52,7 +52,6 @@ jobs:
           cargo ctest --verbose --release
 
       - name: Install into temporary location
-        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
         working-directory: example-project
         run: |
           cargo cinstall --verbose --release --destdir=temp
@@ -81,12 +80,6 @@ jobs:
         run: |
           make
 
-      - name: Install into temporary location
-        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
-        working-directory: example-project
-        run: |
-          cargo cinstall --verbose --release --prefix=/mingw64 --destdir=temp
-
       - name: Install MSYS2
         if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
         uses: msys2/setup-msys2@v2
@@ -97,7 +90,8 @@ jobs:
         if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
         working-directory: example-project
         run: |
-          cp -r temp/mingw64/* /mingw64/
+          # NB: --prefix doesn't seem to matter on Windows
+          cp -r temp/usr/local/* /mingw64/
         shell: msys2 {0}
 
       - name: Test usage from C (using Makefile)


lu-zero (approved):
Overall looks fine, it could be better to use `pkg-config` even on msys2, but can be refined later.

mgeier:
>  it could be better to use `pkg-config` even on msys2

My idea was to test the default paths, so using `pkg-config` would be redundant. Except that MSYS2 doesn't seem to know about the default paths ...

I guess I could use `pkg-config` for now in MSYS2: 7599f2e1e4bccda80404296f3e906d91f72ca7d6

But I would be interested why MSYS2 doesn't know those default paths.
Or does it have different default paths?

1480c1:
Inside msys2, we have multiple roots,
| root | system |
|---|---|
| `/usr` | cygwin |
| `/mingw64` | mingw-w64 gcc 64-bit |
| `/mingw32` | mingw-w64 gcc 32-bit |


and some other experimental ones

mgeier:
Thanks @1480c1 for the info!
So where should I copy the generated library for it to "just work"?

.github/workflows/example-project.yml
```diff
@@ -80,3 +79,23 @@ jobs:
         working-directory: example-project/usage-from-c
         run: |
           make
+
+      - name: Install MSYS2
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        uses: msys2/setup-msys2@v2
+        with:
+          release: false
```
1480c1, line 87:
```suggestion
        with:
          release: false
          install: >-
            mingw-w64-x86_64-pkgconf
            mingw-w64-x86_64-gcc
            base-devel
            git

```
Maybe these packages are needed?

1480c1:
> Thanks @1480c1 for the info!
> So where should I copy the generated library for it to "just work"?

for 64-bit stuff to interact with native windows stuff (not cygwin), you can copy them under `/mingw64`, so a static lib would be under `/mingw64/lib`, a binary and shared dll would be under `/mingw64/bin` and pkg-config files would be under `/mingw64/lib/pkgconfig`

mgeier:
OK, let's try: 2254483720b2029bab0303dcbb03a57a8d0475b0

mgeier:
@1480c1 

So the first attempt didn't work, but in hindsight it makes sense, because when I used the path `/mingw64`, this was still in the default shell and not in the `msys2` shell.

So I tried to fix the path in 55a0ae20d48c801a7605c6c6a5c2272ef52dc785, because the msys2 Action (https://github.com/marketplace/actions/setup-msys2) supposedly uses `C:\msys64` as base directory.

But it still doesn't work ...
What did I do wrong?
Did I use the wrong kind of slashes?

lu-zero:
The makefile is not using pkg-config, it should be fixed to do something along the lines of `pkg-config --cflags --libs example_project`

1480c1:
> So I tried to fix the path in [55a0ae2](https://github.com/lu-zero/cargo-c/commit/55a0ae20d48c801a7605c6c6a5c2272ef52dc785), because the msys2 Action (https://github.com/marketplace/actions/setup-msys2) supposedly uses `C:\msys64` as base directory.

msys2/setup-msys2 actually uses `D:\a\temp` or some other path, the easiest would be to use the `msys2` "shell" as that will auto convert the paths for you

mgeier:
OK, let's try that: 388bab76d239f24a65e915f5bfd25ec64436bd94

However, I think this way the `cargo` command might not be in the PATH.

I think I'll have to do this:

```yml
  - uses: msys2/setup-msys2@v2
    with:
      path-type: inherit
```

... but let's wait for the build to complete ...

@lu-zero 

> The makefile is not using pkg-config

I know, this is intentional.
I'm trying to install the example library to the default system path.

lu-zero:
The default include path for `cc` might not be the right one. probably switching `CC` to the matching `mingw64` one should make it work.

1480c1:
> The default include path for `cc` might not be the right one. probably switching `CC` to the matching `mingw64` one should make it work.

If it gets installed under `/mingw64/include` it should be fine as that's one of the default paths for mingw-w64 gcc (under `/mingw64/bin/gcc`)

mgeier:
OK, thanks, that's what I'm desperately trying in 06c59be80e4cb101e358b75518648f08d6062e2 ... let's see if that works ...

mgeier:
I don't seem to get this to work ... any more ideas?

If not, I'll drop a few commits and go back to the last working state.

mgeier:
I think I finally got it: 0b269472db871b1aa086296f33542ff784bb3a6c!

Thanks for your help @1480c1!

.github/workflows/example-project.yml
```diff
@@ -80,3 +80,29 @@ jobs:
         working-directory: example-project/usage-from-c
         run: |
           make
+
+      - name: Install into temporary location
+        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
+        working-directory: example-project
+        run: |
+          cargo cinstall --verbose --release --prefix=/mingw64 --destdir=temp
```
1480c1, line 88: You may want `shell: msys2 {0}` here as well so `/mingw64` gets expanded properly, unless that's not an issue

mgeier: Thanks, but I tried this and it didn't work.

I *think* the problem is that the string `/mingw64` is supposed to be written into the DLL literally, not in its expanded form.

That's why I added a separate copying step (using the `msys2` shell) where the path is indeed expanded correctly.

But of course I may be totally wrong about this ... I've tried quite a lot of different things ...

1480c1: The main issue I have with the string `/mingw64` being a literal one is that it only makes sense in msys2's stuff.

In our hypothetical test example project, if we were planning to use cargo-c's produced final files outside of msys2 and with other programs, the `/mingw64` string wouldn't make much sense since it's msys2's that is converting the paths to a proper windows path, as without it, `/mingw64` would mean "An absolute path from the root of the current drive." which can change meaning if you are in `D:\>` or `C:\>`, that is assuming the end project is used by something that would read that raw line and interpret it using Window's fopen etc.
If it's using msys2's mingw-w64-x86_64-pkgconf, it will probably be fine (msys2 often uses `sed` to replace the `prefix=` line with `prefix=${MINGW_PREFIX}` is the `prefix=` line is often loaded with an absolute path from GitHub action's CI and using pkgconf will be able to expand that path properly in the end user's layout)

mgeier: I agree that it would look cleaner, but I simply couldn't make it work.

I just tried it again to make sure: https://github.com/mgeier/cargo-c/commit/d12fe0efdba4f76edf94a7f4a3db881325d8f745

Here's the result: https://github.com/mgeier/cargo-c/runs/2512240535

```
cc     run_tests.c  -lexample_project -o run_tests
run_tests.c:1:10: fatal error: example_project.h: No such file or directory
    1 | #include <example_project.h>
      |          ^~~~~~~~~~~~~~~~~~~
```

If you know a way to make this work, please let me know.

But maybe it doesn't matter anyway if the wrong string is embedded in the DLL, I'm not sure if Windows even checks this?

1480c1: try adding in a `command -v cc` and see what it prints out, if it prints out `/usr/bin/cc` or similar, you might need https://github.com/lu-zero/cargo-c/pull/191#discussion_r626675935

mgeier: It prints `/mingw64/bin/cc`, is this good or bad?

1480c1: Should be good, can you try `find /mingw64 -name example_project.h` next? Wasn't it mentioned somewhere needing to pass pkg-config stuff?

mgeier: > `find /mingw64 -name example_project.h`

This returns nothing.

> Wasn't it mentioned somewhere needing to pass pkg-config stuff?

I'm still trying to do all this without using pkg-config (and the current state of this PR runs successfully without needing pkg-config).

1480c1: > > `find /mingw64 -name example_project.h`
> 
> This returns nothing.

Feels like it's not installing to the right place, but oh well, not sure if I can help much more than that 

running
```bash
cargo build
cargo cinstall --prefix=/mingw64
```
and `find /mingw64/ -name example*.h`
returns `/mingw64/include/example_project.h`
so it seems it works on my machine, and running
```bash
make -C usage-from-c
```
compiles just fine
find

mgeier: > Feels like it's not installing to the right place

Is this something that can/should be fixed in `cargo-c`?

> In our hypothetical test example project, if we were planning to use cargo-c's produced final files outside of msys2 and with other programs, the `/mingw64` string wouldn't make much sense

Just out of curiosity I tried this in 7cb6e8e4578d608c07e9d601f54414754bbe7568 where I'm not setting `--prefix` (i.e. it is using the default `/usr/local`), and this seems to work fine nevertheless.

So it looks like this is indeed ignored on Windows (or not even written into the DLL?).

mgeier:
@lu-zero I've added 7cb6e8e4578d608c07e9d601f54414754bbe7568 with a little variation. Feel free to remove this commit if you don't like it.
Both this one and the previous one have successful tests.

lu-zero:
It is good, sorry for not having landed it sooner. Thank you a lot :)

PR #194, opened 2021-05-20
CI: use MSYS2 cargo package for example project

mgeier:
I recently found out that there is a Rust package for MSYS2 (https://packages.msys2.org/base/mingw-w64-rust), and there is even a package for `cargo-c` (https://packages.msys2.org/base/mingw-w64-cargo-c)!

I tried to use the Rust package on CI, but still compile the latest version of `cargo-c` on the fly.

Sadly, there are a few problems ...

Currently, there is a build error during `cargo cinstall`, any ideas how to fix this?

After that, there might still be a problem with the library not being installed into the default location normally used for MSYS2, but for now that's blocked by the compilation error.

~There seems to be another error in the Windows build (the one without MSYS2), which should also be solved, but it isn't really related to this PR.~

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -80,26 +80,42 @@ jobs:
         run: |
           make
 
-      - name: Install MSYS2
-        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
-        uses: msys2/setup-msys2@v2
-        with:
-          release: false
-          update: true
-          install: >-
-            base-devel
+  example-project-msys2:
 
-      - name: Copy installed files
-        if: startsWith(matrix.os, 'windows') && matrix.toolchain-suffix == '-gnu'
-        working-directory: example-project
-        run: |
-          # NB: --prefix doesn't seem to matter on Windows
-          cp -r temp/usr/local/{lib,bin,include} /mingw64/
+    runs-on: windows-latest
+    defaults:
+      run:
         shell: msys2 {0}
 
+    steps:
+      - name: Clone Git repository
+        uses: actions/checkout@v2
+
+      - name: Install MSYS2
+        uses: msys2/setup-msys2@v2
+        with:
+          install: >-
+            make
+            mingw-w64-x86_64-pkg-config
+            mingw-w64-x86_64-rust
+      #      mingw-w64-x86_64-cargo-c
+
+      # NB: cargo-c is available from the package mingw-w64-x86_64-cargo-c
+      # (https://packages.msys2.org/base/mingw-w64-cargo-c),
+      # which is the recommended way to use cargo-c on MSYS2.
+      # For testing purposes, we compile and install the latest version from the Git repo
+      # (which takes much longer than installing the MSYS2 package!):
+
+      - name: Install cargo-c applet
+        run: |
+          cargo install --path .
+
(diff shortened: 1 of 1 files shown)


mgeier:
I've just re-based this PR and the error still occurs:

```
     Running `rustc --crate-name libc 'C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\libc-0.2.98\src\lib.rs' --error-format=json --json=diagnostic-rendered-ansi --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C embed-bitcode=no -Cpanic=abort --cfg 'feature="default"' --cfg 'feature="std"' -C metadata=484b93cfa35492db -C extra-filename=-484b93cfa35492db --out-dir 'D:\a\cargo-c\cargo-c\example-project\target\x86_64-pc-windows-msvc\release\deps' --target x86_64-pc-windows-msvc -L 'dependency=D:\a\cargo-c\cargo-c\example-project\target\x86_64-pc-windows-msvc\release\deps' -L 'dependency=D:\a\cargo-c\cargo-c\example-project\target\release\deps' --cap-lints allow --cfg freebsd11 --cfg libc_priv_mod_use --cfg libc_union --cfg libc_const_size_of --cfg libc_align --cfg libc_core_cvoid --cfg libc_packedN --cfg libc_cfg_target_vendor`
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-pc-windows-msvc` target may not be installed

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: could not compile `libc`
```

mgeier:
I finally managed to make this work!

I had to manually specify `--target` and `--prefix`, which isn't very nice IMHO.
This shouldn't be a blocker for merging this PR, but in the future it would be nice if we could avoid those.

.github/workflows/example-project.yml
```diff
@@ -80,26 +80,42 @@ jobs:
...
+        run: |
+          cargo install --path .
+
+      - name: Install C API for example project
         working-directory: example-project
         run: |
-          # NB: --prefix doesn't seem to matter on Windows
-          cp -r temp/usr/local/{lib,bin,include} /mingw64/
-        shell: msys2 {0}
+          cargo cinstall --verbose --release --prefix /mingw64 --target x86_64-pc-windows-gnu
```
lu-zero, line 116: I guess cargo has the msvc toolchain as default.

mgeier: Yeah, looks like it.

I was hoping to get the right target automatically, since I'm using `cargo` from MSYS2, which should know that it uses the GNU toolchain.

Strangely, when I'm using the MSYS2 package for `cargo-c` (instead of compiling it on-the-fly), the `--target` flag is *not* necessary (see https://github.com/mgeier/cargo-c/runs/4754623622)!

So why does the MSYS2 package of `cargo-c` do the right thing?

But, as I said, I don't think this should be a blocker for now, we can try to figure this out later.

lu-zero:
Thank you!

.github/workflows/example-project.yml
mgeier: I've created a separate issue as a reminder: #244.

mgeier: For completeness, I've also created one about the `--prefix`: #245

PR #198, opened 2021-05-28
Support iOS in install.rs

Diatrus:
Should fix an error I encountered cross compiling rav1e.
```
    Finished release [optimized + debuginfo] target(s) in 35.48s
    Building pkg-config files
    Building header file using cbindgen
thread 'main' panicked at 'not implemented: The target ios- is not supported yet', /Users/diatrus/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-c-0.8.1+cargo-0.53/src/install.rs:94:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
gmake[1]: *** [rav1e.mk:18: rav1e] Error 101
gmake: *** [Makefile:935: rebuild-rav1e] Error 2
```

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -89,7 +89,7 @@ impl LibType {
 
         match (os.as_str(), env.as_str()) {
             ("linux", _) | ("freebsd", _) | ("dragonfly", _) | ("netbsd", _) => LibType::So,
-            ("macos", _) => LibType::Dylib,
+            ("macos", _) | ("ios", _) => LibType::Dylib,
             ("windows", _) => LibType::Windows,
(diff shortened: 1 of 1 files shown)


lu-zero:
Thank you :) Does it produce the library you expect?

Diatrus:
> Thank you :) Does it produce the library you expect?

It does when I use a patched rust to allow me to build cdylibs for iOS. The patch for rust is currently held back, but I figured I'd go ahead and get this part in cargo-c for the future. 👍

lu-zero:
It will appear in the next release :)

PR #200, opened 2021-05-31
Custom compile

lu-zero:
- [x] do not mimic cargo rustc anymore
- [ ] Make a custom `ops::run_tests`


lu-zero:
Landed by cherry-picking.

PR #213, opened 2021-07-05
Do not append .h to the include name if it already has it

lu-zero:


lu-zero:
@sdroege this should fix your problem.

sdroege (approved):

PR #229, opened 2021-10-20
Add datadir support

### LINKED ISSUE #184: Architecture independent install files support

lu-zero:
Fixes #184

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -410,9 +410,10 @@ pub struct LibraryCApiConfig {
     pub rustflags: Vec<String>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct InstallCApiConfig {
     pub include: Vec<InstallTarget>,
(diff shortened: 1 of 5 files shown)


naokiri:
@lu-zero  Yes, this looks really nice. I really appreciate this feature.

lu-zero:
@naokiri did you find time to test it?

naokiri:
Sorry I was busy and couldn't respond for a while.

For my use case I was using other configurations name for directory names and I needed a similar configuration like this PR
https://github.com/lu-zero/cargo-c/pull/237

lu-zero:
Let me land this and then add upon it.

PR #255, opened 2022-04-09
Update to clap-3.1 and cargo-0.61

lu-zero:


sdroege (approved):

lu-zero:
Sadly a design change decision in clap broke silently quite a bit. https://github.com/clap-rs/clap/issues/3621

PR #261, opened 2022-06-04
Add support for Haiku Operating System

Begasus:
For now I've used this to build cargo-c version 0.9.8 (later version contains an error when using latest libgit2-sys), see: https://github.com/haikuports/haikuports/pull/5534
Also have been able to build rav1e 0.5.1 with this.

rav1e

### DIFF
--- a/src/build_targets.rs
+++ b/src/build_targets.rs
@@ -81,7 +81,8 @@ impl BuildTargets {
             | ("freebsd", _)
             | ("dragonfly", _)
             | ("netbsd", _)
-            | ("android", _) => {
+            | ("android", _)
+            | ("haiku", _) => {
                 let static_lib = targetdir.join(&format!("lib{}.a", lib_name));
                 let shared_lib = targetdir.join(&format!("lib{}.so", lib_name));
                 (shared_lib, static_lib, None, None)
--- a/src/target.rs
+++ b/src/target.rs
@@ -72,7 +72,11 @@ impl Target {
         if os == "android" {
             lines.push(format!("-Wl,-soname,lib{}.so", lib_name));
         } else if env != "musl"
-            && (os == "linux" || os == "freebsd" || os == "dragonfly" || os == "netbsd")
+            && (os == "linux"
+                || os == "freebsd"
+                || os == "dragonfly"
+                || os == "netbsd"
+                || os == "haiku")
         {
             lines.push(format!("-Wl,-soname,lib{}.so.{}", lib_name, major));
         } else if os == "macos" || os == "ios" {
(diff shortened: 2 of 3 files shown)


lu-zero:
I'm impressed :) Thank you for the PR!

lu-zero (changes requested):
Looks like you got a couple of stray `)` could you please update your pr please?

src/target.rs
```diff
@@ -75,6 +75,8 @@ impl Target {
             && (os == "linux" || os == "freebsd" || os == "dragonfly" || os == "netbsd")
         {
             lines.push(format!("-Wl,-soname,lib{}.so.{}", lib_name, major));
+        } else if  os == "haiku") {
```
lu-zero, line 78: while at it please fold it in the above branch please.

Begasus: I'll do another round to add it to the above statement today, will get back to that later.

lu-zero: Also please remember to use `cargo fmt` :)

Begasus: > Also please remember to use `cargo fmt` :)

Not sure what you mean by that (sorry not a developer here), looking at "cargo fmt" it would be a separate package (not available in Haiku atm)?
If I look at other recipes around at repology I don't see it being used there?

lu-zero: `cargo fmt` is part of [rustfmt](https://github.com/rust-lang/rustfmt) it is enforced at CI level in many project, including cargo-c.

Begasus: I'll have a look at that later, atm rebuilding with some changes to this PR (specially in `target.rs`)

Begasus: Poking @nielx as he is our rust master :)

Begasus: My bad, seems it's part of our rust package ... https://github.com/haikuports/haikuports/blob/master/dev-lang/rust_bin/rust_bin-1.60.0.recipe#L39

src/build_targets.rs
```diff
@@ -81,7 +81,8 @@ impl BuildTargets {
             | ("freebsd", _)
             | ("dragonfly", _)
             | ("netbsd", _)
-            | ("android", _) => {
+            | ("android", _)
+            | ("haiku", _)=> {
```
lu-zero, line 85:
```suggestion
            | ("haiku", _) => {
```

Otherwise the fmt check would fail.

Begasus: Yeah saw this one earlier, not right, I've run `rust fmt` in the source directory just earlier, will do another round in the morning and see if I can push the changes made by rust for this.

lu-zero (approved):

lu-zero:
Thank you a lot!

Begasus:
Thanks! (also for your patience!) :)

Issue #267, opened 2022-07-22
built shared objects with --library-type cdylib do not have an SONAME on musl


nekopsykose:
(cargo-c 0.9.11)  

building a library (libimagequant in this case) with:  

```
	cargo cbuild --release \
		--prefix /usr \
		--library-type cdylib

	cargo cinstall --release \
		--prefix /usr \
		--destdir "$pkgdir" \
		--library-type cdylib
```

(like [here](https://git.alpinelinux.org/aports/tree/community/libimagequant/APKBUILD?id=41dc5dcc7799046787d37e0c057851427fe67d21))
(cdylib only needed because of #180)  

then generated a library without an SONAME:  

```
$ readelf -d /usr/lib/libimagequant.so.0 
Dynamic section at offset 0x74920 contains 24 entries:
  Tag        Type                         Name/Value
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../../../../../../../../../../../../usr/lib/rustlib/x86_64-alpine-linux-musl/lib]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.musl-x86_64.so.1]
 0x000000000000001e (FLAGS)              ORIGIN BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW ORIGIN
 0x0000000000000007 (RELA)               0x1830
 0x0000000000000008 (RELASZ)             24816 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffff9 (RELACOUNT)          967
 0x0000000000000017 (JMPREL)             0x7920
 0x0000000000000002 (PLTRELSZ)           120 (bytes)
 0x0000000000000003 (PLTGOT)             0x771f8
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000006 (SYMTAB)             0x270
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0xfe4
 0x000000000000000a (STRSZ)              2122 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0xe50
 0x000000000000000c (INIT)               0x1cb6c
 0x000000000000000d (FINI)               0x1cb79
 0x000000006ffffff0 (VERSYM)             0xd38
 0x000000006ffffffe (VERNEED)            0xe20
[...]

Issue #180, opened 2021-04-02
Shared libraries are not built on Alpine Linux


OTLabs:
Tried to build rav1e 0.4.1 with cargo-c 0.8.0 on Alpine Linux and got no any .so libraries built. I was expecting to get: librav1e.so.0 and librav1e.so.0.4.1.

rav1e 0.4.0 with cargo-c 0.7.1 build them without any problem.

Actually it is not a problem,  I am just curious to find out the origin of this change in behavior and to know what to expect in future builds as I used to have a separate package for shared libraries.

lu-zero:
Alpine Linux uses musl and the musl target in rust is assumed to require static-only builds.

https://github.com/lu-zero/cargo-c/commit/3dc0c1e66ae8435449f8e650a0ed2ccdad3f9ffe always set the target
and https://github.com/lu-zero/cargo-c/commit/736a84292d4d90849cdf71621f8caee49c3dff22 defaults to static-only if not otherwise stated. 

In theory the newer rustc (maybe still nightly) would have a specific target for musl+shared-library that would make the default work as you expect. For now passing `--library-type staticlib,cdylib` should produce the .so on Alpine.

lu-zero:
See here https://github.com/rust-lang/rust/pull/82556 for the current status.

OTLabs:
I have tried `--library-type staticlib,cdylib` but got following error:

```
692 error: 'staticlib,cdylib' isn't a valid value for '--library-type <LIBRARY-TYPE>...'
693	[possible values: cdylib, staticlib]
694	Did you mean 'staticlib'?
695USAGE:
696    cargo-cbuild cbuild --frozen --library-type <LIBRARY-TYPE>... --prefix <prefix> --release
```

Looks like I could not use both arguments same time, right?

lu-zero:
`--library-type staticlib --library-type cdylib` sorry.

OTLabs:
Thank you for a tip, checking now the build...

OTLabs:
Works great! Thank you so much for you kind attention and help!

lu-zero:
Once cargo is able to tell the difference between musl-with-dylibs and musl-without it will work out of box. :)

richfelker:
This is a mistreatment of the musl target we're trying to get fixed in upstream Rust. Like any other platform with dynamic linking, linux-musl target should produce dynamic binaries by default and support the production of shared libraries, and only do static linking when a static-linked binary is requested.

[...]

PR #268, opened 2022-07-22
Add sonames on musl

nekopsykose:
the musl platform also supports shared objects having an SONAME, and
this function is called only when making a 'shared object', so this
should always be added

fixes https://github.com/lu-zero/cargo-c/commit/04a92bc497698cc822bc969307744eb929cc1eab where this was removed
even when generating a dynamic library forcefully by setting library
type to cdylib

fixes https://github.com/lu-zero/cargo-c/issues/267

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -71,12 +71,11 @@ impl Target {
 
         if os == "android" {
             lines.push(format!("-Wl,-soname,lib{}.so", lib_name));
-        } else if env != "musl"
-            && (os == "linux"
-                || os == "freebsd"
-                || os == "dragonfly"
-                || os == "netbsd"
-                || os == "haiku")
+        } else if os == "linux"
+            || os == "freebsd"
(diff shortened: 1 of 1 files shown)


nekopsykose:
reverting all of 04a92bc497698cc822bc969307744eb929cc1eab instead does give warnings with the default -unknown-linux-musl toolchains:  

```
warning: dropping unsupported crate type `cdylib` for target `x86_64-unknown-linux-musl`
```

(though they are of course harmless)  

but when staticlib is used, this code path is entirely unused (it's called `shared_object_link_args` and only gets called when making a cdylib), so when someone does have a musl toolchain that can make cdylibs, the soname should be there; this conditional was always broken

nekopsykose:
if there was a way to detect 'target supports cdylib' we could fix #180 too, but i don't think there is a way

lu-zero:
Please run rustfmt on your commits :)

PR #270, opened 2022-08-06
Fix typo on readme

naokiri:
Change of 'assets' to 'asset' is based on the code below, and my trial and error on my project. 
https://github.com/lu-zero/cargo-c/blob/d8f85933579911a4365a494a63a3496307a6651d/src/build.rs#L692


lu-zero (approved):
Thank you!

PR #272, opened 2022-08-09
Install the windows DLL in the libdir if it is custom

### LINKED ISSUE #271: Libraries that aren't in libdir should not be moved to bindir on Windows
Currently if you [specify an `install_subdir` for a library](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/d240bbc4e24d07dac1551be292e5e75794d00096/audio/audiofx/Cargo.toml#L49), on Windows the `.lib`, `.dll.lib` and `.def` will be installed into `$prefix/lib/$subdir` but the `.dll` will be installed into `$prefix/bin/$subdir`.

This is surprising, and also not wanted when building, for example, plugins. You only want to do that for DLLs that applications link to, and hence might load from PATH at runtime. This should only happen with import libraries that are in libdir or other directories in the linker path.

In Meson, we only do this automatic transporting of the DLL if the library is installed into `$libdir`. Maybe cargo-c can do the same? It's a good heuristic and we've had no complaints about it.

lu-zero:
We assume they are plugins as meson does, as discussed in #271.

Fixes #271

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -243,11 +243,18 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
                     lib.install(capi_config, shared_lib, &install_path_lib)?;
                 }
                 LibType::Windows => {
-                    let install_path_bin = append_to_destdir(destdir.as_deref(), &paths.bindir);
-                    create_dir_all(&install_path_bin)?;
-
                     let lib_name = shared_lib.file_name().unwrap();
-                    copy(shared_lib, install_path_bin.join(lib_name))?;
+
+                    if capi_config.library.install_subdir.is_none() {
+                        let install_path_bin = append_to_destdir(destdir.as_deref(), &paths.bindir);
+                        create_dir_all(&install_path_bin)?;
+
+                        copy(shared_lib, install_path_bin.join(lib_name))?;
+                    } else {
+                        // We assume they are plugins, install them in the custom libdir path
+                        copy(shared_lib, install_path_lib.join(lib_name))?;
+                    }
+
                     let impl_lib = build_targets.impl_lib.as_ref().unwrap();
                     let impl_lib_name = impl_lib.file_name().unwrap();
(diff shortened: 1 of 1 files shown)


src/install.rs
```diff
@@ -243,17 +243,23 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
...
+                        copy(shared_lib, install_path_bin.join(lib_name))?;
+                        let impl_lib = build_targets.impl_lib.as_ref().unwrap();
+                        let impl_lib_name = impl_lib.file_name().unwrap();
+                        copy(impl_lib, install_path_lib.join(impl_lib_name))?;
+                        let def = build_targets.def.as_ref().unwrap();
+                        let def_name = def.file_name().unwrap();
+                        copy(def, install_path_lib.join(def_name))?;
+                    } else {
+                        // We assume they are plugins, install them in the custom libdir path
+                        copy(shared_lib, install_path_lib.join(lib_name))?;
```
nirbheek, line 261: If you want to emulate Meson behaviour, you will want to copy the implib and the def file to the same location.

Basically, the only thing that would change if the install path is custom, is that shared_lib will be installed to `install_path_lib` instead of `install_path_bin`

sdroege:
It probably makes sense to add a big comment somewhere in the code that this is Windows-only logic. AFAIU this would cause problems if libdir is provided manually to install into `/usr/lib/x86_64-linux-gnu` instead of `/usr/lib`, for example, and the same logic is used for some other future feature that is also used on Linux.

src/install.rs
lu-zero: Ah, even simpler then :)

lu-zero:
That changes libdir, doesn't add to libdir as the option you set in Cargo.toml.

nirbheek:
LGTM!

lu-zero:
Please test as much as possible that the new release will happen in probably few weeks ^^

PR #277, opened 2022-09-01
Fix linking with Android NDK linker

nirbheek:
Android NDK `ld.gold` cannot handle `-Wl`, and it's incorrect anyway since all the other linker arguments passed by cargo do not have a `-Wl` prefix.

```
<lots of args snipped> "--gc-sections" "-shared" "-zrelro" "-znow" "-O1" "-Wl,-soname,libgsturiplaylistbin.so"
   = note: arm-linux-androideabi-ld.gold: -Wl,-soname,libgsturiplaylistbin.so: unknown option
           arm-linux-androideabi-ld.gold: use the --help option for usage information
```

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -70,7 +70,7 @@ impl Target {
         let env = &self.env;
 
         if os == "android" {
-            lines.push(format!("-Wl,-soname,lib{}.so", lib_name));
+            lines.push(format!("-soname,lib{}.so", lib_name));
         } else if os == "linux"
             || os == "freebsd"
             || os == "dragonfly"


lu-zero:
Thank you for catching it :)

lu-zero (commented):
Android doesn't use llvm for that yet though ?

nirbheek:
> Android doesn't use llvm for that yet though ?

This is with Android NDK 21, maybe the newer NDKs use lld! We haven't updated the NDK yet. Using the right linker requires cargo configuration anyway, you need to create `.cargo/config.toml` with the following contents:

```
[target.<triple>]
linker = "/path/to/linker"
rustflags = ["-L/path/to/libdir"]
```

This is also not something cargo-c can detect automatically, so nothing for us to do here.

lu-zero:
the linker is supposed to be the C compiler...

nirbheek:
I already tried that.

First, `arm-linux-androideabi-gcc` doesn't exist. It's clang now with a bunch of arguments to select the right arch. That is broken by default with cargo because `target.<triple>.linker` only takes a string, not an array of strings.

With a hacky shell script you can make it work, but then it fails differently because rustc is passing arguments without `-Wl` to it:

```
<lots of args snipped> "--gc-sections" "-shared" "-zrelro" "-znow" "-O1" "-Wl,-soname,libgstrsregex.so"
  = note: clang: error: unsupported option '--version-script=/tmp/rustcgrBvXq/list'
          clang: error: unsupported option '--as-needed'
          clang: error: unsupported option '--start-group'
          clang: error: unsupported option '--end-group'
          clang: error: unsupported option '--eh-frame-hdr'
          clang: error: unsupported option '--gc-sections'
          clang: error: unknown argument: '-znoexecstack'
          clang: error: unknown argument: '-zrelro'
          clang: error: unknown argument: '-znow'
```

lu-zero:
I tried and:
- google had the brilliant idea to remove libgcc in newer ndk
- the "fix" in nightly apparently does not work at all
- https://github.com/godot-rust/godot-rust/pull/920/files this seems the sanest workaround

lu-zero:
Using
```
[target.aarch64-linux-android]
linker = "/opt/android/sdk/ndk/25.1.8937393/toolchains/llvm/prebuilt/darwin-x86_64/bin/aarch64-linux-android33-clang"
```

with the above workaround makes the example build fine with
```
cargo cbuild --manifest-path=example-project/Cargo.toml --target aarch64-linux-android
```

lu-zero:
If you feel inclined you may port https://github.com/bbqsrc/cargo-ndk/pull/67/files to cargo-c ^^;

nirbheek:
That would probably work with Android NDK 25, but I am using NDK 21. It's understandable if you don't want to support NDK 21. I'll close the PR :)

lu-zero:
It should work exactly the same for NDK 21, let's figure out on irc the details :)

nirbheek:
@lu-zero which IRC channel and server?

lu-zero:
libera and #rust-av would be fine, or poke me on query :)

sdroege:
We also dropped this patch now because it was only necessary for pre-25 NDK with ld.gold. NDK 25 with the LLVM linker doesn't have any such problems.

PR #282, opened 2022-09-23
add support for illumos

hadfl:
With this change in place I was able to build the `rav1e` C-API on [OmniOS](https://omnios.org/) (an illumos based OS).


lu-zero:
Thank you!

PR #289, opened 2022-11-10
ci(deploy): Use musl abi for aarch64 linux binary, add i686 linux

fdintino:
I'm using cross to build a binary for the aarch64-unknown-linux-musl target, which avoids glibc ABI compatibility issues with the gcc from crossbuild-essential-arm64 and so can run on more platforms. I've also added i686 linux musl binaries to the deploy workflow. The binary created by aarch64-unknown-linux-gnu only runs on the most recent distributions because of the version of glibc it is linked against. Executing the binaries from the v0.9.14-deploy release on debian buster (and el7) fails with the error:

```
cargo-cinstall: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by cargo-cinstall)
cargo-cinstall: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by cargo-cinstall)
```

I've created a test release on my fork [here](https://github.com/fdintino/cargo-c/releases/tag/v0.9.14-musl-aarch64-i686). I've also written a workflow job to test that the binaries can run against debian buster, but omitted it from this PR because I wasn't sure it was worth the added complexity. Still, you can see it in [this commit](https://github.com/fdintino/cargo-c/commit/97c15b167658cf8ab657e9dffacea2e4a6251979), and its run against the binaries built from this PR [here](https://github.com/fdintino/cargo-c/actions/runs/3436562324).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,19 +32,11 @@ jobs:
 
     - name: Build cargo-c
       run: |
-        cargo build --release
-
-    - name: Strip gnu binaries
-      if: matrix.conf == 'gnu'
-      run: |
-        strip "target\release\cargo-capi.exe" `
-              "target\release\cargo-cbuild.exe" `
-              "target\release\cargo-cinstall.exe"
-
+        cargo build --profile release-strip
 
     - name: Create zip
       run: |
-        cd target/release
+        cd target/release-strip
         7z a ../../cargo-c-windows-${{ matrix.conf }}.zip `
              "cargo-capi.exe" `
              "cargo-cbuild.exe" `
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -58,43 +58,31 @@ jobs:
 
   linux-binaries:
     strategy:
+      fail-fast: false
       matrix:
         target:
+         - i686-unknown-linux-musl
```
lu-zero, line 64: this is maybe spurious, I guess? I guess also ppc could target musl for uniformity

lu-zero:
Thank you for the submission :) I left two comments, but the whole idea of using cross is great.

.github/workflows/deploy.yml
fdintino: I'm not sure I understand your meaning. This is the build for i686 architectures; it's still commonplace to release 32-bit python wheels, which is my specific use case as the maintainer of [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin).

I had tried to target powerpc64le-unknown-linux-musl, but it's not trivial because that target is a [tier 3 platform](https://doc.rust-lang.org/rustc/platform-support.html#tier-3), which means that there aren't any official builds for the toolchains. So we would have to do that as part of this workflow.

fdintino: It is possible to build ppc musl if it is run in a rust-musl-cross container. See https://github.com/fdintino/cargo-c/commit/b50119686e058eea775c9f8f0656a6b892ecbfed

lu-zero: Too much complexity for little gain, but thank you a lot for figuring out how to do it. :)

Issue #295, opened 2022-12-07
C appli fails to link with a RS lib built with versioning = false


stephan57160:
Created a small C testing tool using a small RUST lib, built with cargo-c
`cargo-c` is the last official crate version.

I configure Cargo.toml with `versioning = true`:
* both build fine
* the C testing tool is able to link with the RS lib.

But when `versioning = false`, the C testing tool link fails with:
```
xxx: error while loading shared libraries: libxxxx.so.0: cannot open shared object file: No such file or directory
make[1]: *** [Makefile:43 : all] Erreur 127
```

As there is no `libxxxx.so.0` at all, I ran `readelf`:
```
stephan@PC2-Stephan:~/git/libreceipt-rs$ readelf -d production/linux/lib/libreceiptrs.so 

Dynamic section at offset 0x47808 contains 29 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000e (SONAME)             Library soname: [libxxxx.so.0]           # <-- has a SO version !!
 0x000000000000000c (INIT)               0x5000
 ```

With `pathelf`, if I replace the SONAME with `libxxxx.so`, the link succeeds.

lu-zero:
I guess we should document better the expectation, making so if versioning is false the link line uses the plain libname.so is easy but if you used it to install the library in the system you are in for lots of fun surprises =/

stephan57160:
> if you used it to install the library in the system you are in for lots of fun surprises =/
Not soon, but one day ... maybe.

No doubt that, "fun surprises" will bring you fun as well :-)

lu-zero:
Not really, `versioning = false` expects you to not install the library as a system one, trying to do that is unsupported and would subtly break anything using such library.

Probably we could rename it to `plugin` or such so the intention is clear.

stephan57160:
I can try that tomorrow ... at least, just to have some fun :-)

lu-zero:
You stated how the breakage happens, I'd rather not enable people to do something wrong if I can.

sdroege:
See https://github.com/lu-zero/cargo-c/pull/296

stephan57160:
Thx

eli-schwartz:
[...]

PR #296, opened 2022-12-11
Don't include version in soname ELF header if versioning is disabled

sdroege:
And the equivalent change for macOS/iOS.

Fixes https://github.com/lu-zero/cargo-c/issues/295

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -78,15 +78,27 @@ impl Target {
             || os == "haiku"
             || os == "illumos"
         {
-            lines.push(format!("-Wl,-soname,lib{}.so.{}", lib_name, major));
-        } else if os == "macos" || os == "ios" {
-            let install_ver = if major == 0 {
-                format!("{}.{}", major, minor)
+            if capi_config.library.versioning {
+                lines.push(format!("-Wl,-soname,lib{}.so.{}", lib_name, major));
             } else {
-                format!("{}", major)
+                lines.push(format!("-Wl,-soname,lib{}.so", lib_name));
+            }
+        } else if os == "macos" || os == "ios" {
+            let line = if capi_config.library.versioning {
+                let install_ver = if major == 0 {
+                    format!("{}.{}", major, minor)
+                } else {
+                    format!("{}", major)
+                };
+                format!("-Wl,-install_name,{1}/lib{0}.{5}.dylib,-current_version,{2}.{3}.{4},-compatibility_version,{5}",
+                        lib_name, libdir.display(), major, minor, patch, install_ver)
+            } else {
+                format!(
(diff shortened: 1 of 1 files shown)


src/target.rs
```diff
@@ -78,15 +78,27 @@ impl Target {
             || os == "haiku"
```
sdroege, line 78: Some lines above, why do we never set a version in the soname for Android @lu-zero ?

src/target.rs
```diff
@@ -78,15 +78,27 @@ impl Target {
...
+                    format!("{}.{}", major, minor)
+                } else {
+                    format!("{}", major)
+                };
+                format!("-Wl,-install_name,{1}/lib{0}.{5}.dylib,-current_version,{2}.{3}.{4},-compatibility_version,{5}",
+                        lib_name, libdir.display(), major, minor, patch, install_ver)
             } else {
-                format!("{}", major)
+                format!(
+                    "-Wl,-install_name,{1}/lib{0}.dylib",
```
sdroege, line 97: @nirbheek Does this match what meson is doing on macOS?

src/target.rs
lu-zero: because android does not have the version it its libraries.

sdroege: First time I hear that :) Do you have some more details about that?

lu-zero: It is a well known annoyance, you may look at stackoverflow or the hopefully now closed bugs at common software such openssl ^^; I cannot find a document that explicitly mentions it but that's how it is, assuming nothing changed in the past year.

stephan57160:
Thx all.

Any idea when this will come "official" ?

src/target.rs
sdroege: How strange :) Thanks, I learned something new.

stephan57160: @lu-zero: This is still true with the last NDKs. 
does not have and does not want !

src/target.rs
nirbheek: Yes, it's close enough to that.

On macOS, Meson allows you to set `current_version` and `compatibility_version` specifically with the [`darwin_versions:` kwarg](https://mesonbuild.com/Reference-manual_functions.html#shared_library). If those are not specified, the default is to use the soname version for both (`soversion:` kwarg). If the `soversion:` kwarg is not specified, the first component of the project version is picked (and this trickles down to everything derived from `soversion`).

`-install_name` is always `libfoo.${soversion}.dylib`, same things as above apply.

sdroege:
@lu-zero I guess let's get this in then and a new release with these changes would be useful to :)

sdroege:
Let's get this in then? :)

PR #298, opened 2022-12-17
macos: Allow larger install_name entries

nirbheek:
This is needed to be able to add multiple RPATH entries, and also to change the install_name itself. This is enabled by default in Meson.

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -99,7 +99,9 @@ impl Target {
                     libdir.display()
                 )
             };
-            lines.push(line)
+            lines.push(line);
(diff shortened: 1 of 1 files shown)


lu-zero:
I assume the option is supported by all the linkers we care about.

lu-zero (approved):

nirbheek:
Hm, the clippy job is not smart enough to know which of these warnings are caused by the changes in this MR (none). I'll add a commit that fixed the warnings I guess.

lu-zero:
Just rebase and you'll be fine, hopefully :)

PR #299, opened 2022-12-18
Update to cargo-0.67 and clap-4

lu-zero:
Notes:
- the help is fairly different, we might disable the colors while we wait for https://github.com/clap-rs/clap/issues/3234 since the current default is a tad too red for my taste.
image
- The changes are many I need help from the usual suspects to make sure I do not break anything (hi @slomo and @nirbheek)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-c"
-version = "0.9.14+cargo-0.66"
+version = "0.9.14+cargo-0.67"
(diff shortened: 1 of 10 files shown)


sdroege:
I'll test this with the GStreamer plugins and ebur128 once the CI here passes. The changes look good to me though :)

lu-zero:
There was a OsString vs String in one place, probably there is another prowling about ^^;

sdroege:
Looks good in my testing :)

Issue #280, opened 2022-09-05
Consider naming DLLs built with MinGW as `libfoo.dll`


nirbheek:
DLLs built with Autotools (which uses MinGW of course) are named `libfoo.dll`, and other build systems on Windows that use MSVC (like Visual Studio project files or CMake) use `foo.dll`.

So this is a convenient way to differentiate between DLLs built with MSVC and those built with MinGW, which is convenient for projects that provide binaries built with both toolchains. This is also what Meson does.

Would you like to follow Meson in this or do you prefer sticking with Cargo defaults as to which target platforms prefix `lib` to the library name?

lu-zero:
I think would be nice to give the users the choice, that way it should be less confusing. Possibly we could tie more settings to a `--meson` cli option.

nirbheek:
Yeah there's other behaviours also that Meson has, such as [always naming static libraries `libfoo.a` even when built with MSVC](https://mesonbuild.com/FAQ.html#why-does-building-my-project-with-msvc-output-static-libraries-called-libfooa) and import libraries are always `foo.lib` so that `-lfoo` works in pkg-config files with GCC, off the top of my head.

AFAICT cargo-c uses `foo.lib` for static archives and `foo.dll.lib` for import libraries?

lu-zero:
We can double check, windows is not a platform I use.

lu-zero:
Is this still a problem?

nirbheek:
Yes. We hack around it by [renaming the files after install](https://gitlab.freedesktop.org/gstreamer/cerbero/-/blob/74b9ebcea7c4f3d457f610b7e6bc5eb6d80f5094/recipes/gst-plugins-rs.recipe#L142-154).

lu-zero:
I guess it is a good time to add this additional bit of logic then

lu-zero:
@nirbheek do you have time to help me on this?

amyspark:
Hey @lu-zero , @nirbheek just sent this over to me. I'll be taking a look into it later today.

lu-zero:
Great! there are a couple of meson-parity issues that I keep postponing since I do not have access to windows... Feel free to message on irc in case.

PR #300, opened 2023-01-03
Prefix import libraries with `lib` on mingw

MehdiChinoune:


sdroege:
This is kind of related to https://github.com/lu-zero/cargo-c/issues/280, but only doing part of that issue. Can you update it accordingly?

CC @nirbheek

MehdiChinoune:
> Can you update it accordingly?

update what?

src/build.rs
```diff
@@ -232,7 +232,7 @@ fn build_implib_file(
             dlltool_command.arg("-D").arg(format!("{}.dll", name));
```
kmilos, line 232: Prefix also the DLL to match Autotools/CMake/Meson behaviour?

src/build_targets.rs
```diff
@@ -103,7 +103,7 @@ impl BuildTargets {
                 let impl_lib = if env == "msvc" {
                     targetdir.join(format!("{}.dll.lib", lib_name))
                 } else {
-                    targetdir.join(format!("{}.dll.a", lib_name))
+                    targetdir.join(format!("lib{}.dll.a", lib_name))
```
kmilos, line 106: Also prefix depending on msvc or not on line 102 above?

src/build.rs
MehdiChinoune: I tried to, but for some reason It failed to find the dll file when installing.

kmilos: Perhaps because it needs to be paired w/ the change below as well?

src/build_targets.rs
MehdiChinoune: What do you?
If env==msvc import library has an output name `${libname}.dll.lib`, else (==gnu) it has `lib${libname}.dll.a`

src/build.rs
MehdiChinoune: I tried that too, You could try it yourself. Patch cargo-c As I did and see if it works for you.

src/build_targets.rs
kmilos: Not only `imp_lib` here, but also `shared_lib` on line 102 above

kmilos:
@MehdiChinoune Indeed, `{}.dll` is produced no matter what, as if cargo itself needs to be patched...

kmilos:
Ah, perhaps also adding to https://github.com/lu-zero/cargo-c/blob/7760ca3d4cb31a7d8a3cbfe003576e2f47acd7f7/src/target.rs#L107-L110

PR #301, opened 2023-01-03
support aarch64 on mingw

MehdiChinoune:


lu-zero:
Thank you, can you please prepare a patch to add a CI item to try to test it?

MehdiChinoune:
> Thank you, can you please prepare a patch to add a CI item to try to test it?

GH does not offer any aarch64 virtual environments.
We -at MSYS2- rely on one only donated machine to build packages.
The patch was applied and we tested it to build rav1e on aarch64.

Issue #302, opened 2023-01-11
`cargo cinstall` vs. `cargo install`: `--debug` requires an argument?


jbeich:
Looking at dbf1d645323e I don't understand what should be passed as an argument. A profile name e.g., `dev`?

```
$ cargo cbuild --version
cargo-c 0.9.15+cargo-0.67

$ cargo install -h 2>&1 | fgrep debug
      --debug                   Build in debug mode (with the 'dev' profile) instead of release mode
$ cargo cinstall -h 2>&1 | fgrep debug
      --debug <debug>                Build in debug mode instead of release mode
```

lu-zero:
It seems a bug. Thank you for spotting it :)

lu-zero:
It will be released with the new cargo if you do not need it before. Otherwise I'll prepare a point release this weekend.

PR #303, opened 2023-01-12
cli: drop bogus arguments after cargo update

jbeich:
Fixes https://github.com/lu-zero/cargo-c/issues/302


lu-zero:
Thank you a lot!

lu-zero (approved):

PR #320, opened 2023-05-19
Add a library.import_library configuration option

lu-zero:
It is used to disable the generation of import_libraries when unnecessary in platforms that need it to link libraries (e.g. Windows).

@sdroege could you please check if it would work for gst?

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,9 @@ versioning = false
 # Add `-Cpanic=abort` to the RUSTFLAGS automatically, it may be useful in case
 # something might panic in the crates used by the library.
 rustflags = "-Cpanic=abort"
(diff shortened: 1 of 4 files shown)


sdroege:
From reading the code, that should work fine, yes.

lu-zero:
@realmazharhussain It should address your question, but I'd like some volunteers to try it and I do not have a windows box to try myself.

realmazharhussain:
Sorry for being late. I tried it on msys2 in UCRT64 environment and it seems to be working as intended.

lu-zero:
If you do not need it sooner it will appear in the next release then :)

realmazharhussain:
There is no hurry.

PR #330, opened 2023-07-31
README: add ebur128 to Users section

jbeich:
See also https://github.com/freebsd/freebsd-ports/commit/4cd440845e24


jbeich:
Looking at cargo-c consumers in Alpine Ports (aports) it may also worth adding libdovi to Users section.

lu-zero:
Thank you :)

lu-zero (approved):

Issue #345, opened 2023-09-30
More granular SONAME versioning


kpcyrd:
Related to https://github.com/rustls/rustls-ffi/issues/345 and https://github.com/rustls/rustls-ffi/pull/274.

I've tried to use cargo-c together with Arch Linux packaging tools, which is able to automatically detect and embed the SONAME into package metadata, but noticed I get this:

```
provides = librustls.so=0-64
```

After looking into how this works I noticed it's using a string embedded in the .so file:

```
% readelf -d /usr/lib/librustls.so | rg Lib
 0x000000000000000e (SONAME)             Library soname: [librustls.so.0]
```

I found the relevant part in the cargo-c codebase:

https://github.com/lu-zero/cargo-c/blob/0819defe7fcfcb26156ed9c439b433677d48dbac/src/target.rs#L81-L85

There are two things about this:

1) if the version string is `0.11.0`, I think the soname should be `librustls.so.0.11`
2) even if the version string is `>= 1.0.0`, I think it should be possible to configure a major.minor (or even major.minor.patch) SOVER

Specifically with librustls I would like to be able to build it with `-Wl,-soname,librustls.so.0.11.0` until the developers are comfortable making ABI commitments. :)

lu-zero:
You are right, we'll have a new cargo soon and I'd release your improvement/fix with it if is fine for you. If you prefer having it in a point release please tell me and I'll prepare it :)

kpcyrd:
Sounds great, thank you! :) Looking forward to the next cargo release.

lu-zero:
The new cargo is out, let me know if everything works as you need.

radhermit:
This appears to break the default case where the soname version isn't specified in metadata.

For example, a library I'm building with cargo-c now has "*.so.0.0" instead of "*.so.0" for the SONAME and since cargo-c doesn't create the related *.so.0.0 file it breaks all the bindings that try to link to the nonexistent library.

lu-zero:
in your case the soversion should be `lib{lib_name}.so.{major}.{minor}.{patch}`, definitely more to fix.

radhermit:
That makes sense, I didn't revert to the previous version to verify my guess.

lu-zero:
I'll try to refactor and cleanup a bit more since there is one more corner case:

| version | links | soversion |
|- |- |- |
| 0.0.n | libfoo.so | libfoo.so.0.0.n |
[...]

PR #346, opened 2023-09-30
Fix sover detection for 0-based versioning

kpcyrd:
Related to #345

With this patch applied Arch Linux generates this metadata (emphasis on `Provides`):

```
Name            : librustls
Version         : 0.11.0-1
Description     : Use rustls from languages other than Rust
Architecture    : x86_64
URL             : https://github.com/rustls/rustls-ffi
Licenses        : Apache-2.0  MIT
Groups          : None
Provides        : librustls.so=0.11-64
```

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -69,6 +69,12 @@ impl Target {
         let os = &self.os;
         let env = &self.env;
 
+        let sover = if major == 0 {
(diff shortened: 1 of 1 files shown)


lu-zero:
Sounds good, thank you :)

PR #350, opened 2023-10-10
Complete the soversion granularity

lu-zero:


lu-zero:
@kpcyrd could you please test and confirm that works for your purpose as well?

src/lib.rs
```diff
@@ -5,3 +5,22 @@ pub mod config;
...
+    fn main_version(&self) -> String {
+        if self.major == 0 {
+            if self.minor == 0 {
+                format!("{}.{}.{}", self.major, self.minor, self.patch)
+            } else {
+                format!("{}.{}", self.major, self.minor)
+            }
+        } else {
+            format!("{}", self.major)
+        }
```
kpcyrd, line 24:
```suggestion
        match (self.major, self.minor, self.patch) {
            (0, 0, patch) => format!("0.0.{patch}"),
            (0, minor, _) => format!("0.{minor}"),
            (major, _, _) => format!("{major}"),
        }
```

src/install.rs
```diff
@@ -146,12 +144,14 @@ impl UnixLibNames {
...
-            .arg(&self.with_full_ver)
-            .arg(install_path_lib.join(&self.with_major_ver));
-        let _ = ln_sf.status().unwrap();
+        if self.with_main_ver != self.with_full_ver {
+            let mut ln_sf = std::process::Command::new("ln");
+            ln_sf.arg("-sf");
+            ln_sf
+                .arg(&self.with_full_ver)
+                .arg(install_path_lib.join(&self.with_main_ver));
+            let _ = ln_sf.status().unwrap();
```
kpcyrd, line 153: This panics if the `ln` binary couldn't be executed, and ignores the `ExitStatus` if `ln` returned an error (below for `self.canonical` too).

src/lib.rs
lu-zero: It is a nice idea.

lu-zero: Did you test that it works for you though? :)

PR #352, opened 2023-11-11
Add version_suffix_components setting

### LINKED ISSUE #345: More granular SONAME versioning

kpcyrd:
Related to #345 (this patch is based on my comment there) and rustls/rustls-ffi/pull/274.

This adds a setting to disable automatic semver SONAMEs and to pick an arbitrary number of components from the version string for the SONAME.

This feature allows two additional use-cases for library developers to pick from:

## `.so.0`, `.so.1`, `.so.2`, ...

For a C-style breaking-change counter use a config like this:

```toml
[package.metadata.capi.library]
name = "imagequant"
version = "0.0.0"
version_suffix_components = 1
```

Increase the major version for every time the ABI is broken:

```toml
[package.metadata.capi.library]
name = "imagequant"
version = "1.0.0"
version_suffix_components = 1
```

This pattern is already supported, but one has to start at `.so.1` because it's currently not possible to configure `.so.0`.

Doing this does not interfer with pkg-config metadata:

```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: imagequant
Description: Convert 24/32-bit images to 8-bit palette with alpha channel.
Version: 4.0.4
Libs: -L${libdir} -limagequant
Cflags: -I${includedir}
Libs.private:  -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc
```

## Opt-out of a stable ABI

To read the version from `[package.version]` but opt-out of a stable ABI:

```toml
[package.metadata.capi.library]
name = "rustls"
version_suffix_components = 3
rustflags = "-Cmetadata=rustls-ffi"
```

"Why not do static linking at this point?" - doing it like this has 3 effects:

- Binary code is available from a single shared object in the dynamic linking ecosystem and does not need to be duplicated into every consuming executable
- Binary code can be centrally security patched by downstream instead of having to replace embedded copies in every consuming executable
- For every new upstream release, all consuming executables need to be rebuilt because the SONAME is always going to change, intentionally preventing the new binary code to be loaded by an executable that assumes the ABI of a previous version

Another way of opting out of a stable ABI is "use 0.1.0 and only ever bump the minor version". cargo-c default settings would then make sure every release is in it's own `libfoo.so.0.1`, `libfoo.so.0.2` SONAME namespace, until comfortable with making ABI compatible `0.X.{patch}` releases that still provide the first two effects, but do not require a rebuild of consuming executables.

---

I originally named this setting `suffix_components` but I'm indecisive which name is better.

As far as I can tell this is the last feature needed to close #345 as completed.

cc: @lu-zero @cpu @kornelski @kyrias

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -14,6 +14,7 @@ use cargo::util::command_prelude::{ArgMatches, ArgMatchesExt, CompileMode, Profi
 use cargo::util::interning::InternedString;
 use cargo::{CliResult, Config};
 
+use anyhow::Context as _;
 use cargo_util::paths::{copy, create, create_dir_all, open, read, read_bytes, write};
 use semver::Version;
 
@@ -406,8 +407,31 @@ pub struct LibraryCApiConfig {
     pub version: Version,
     pub install_subdir: Option<String>,
     pub versioning: bool,
+    pub version_suffix_components: Option<u8>,
     pub import_library: bool,
     pub rustflags: Vec<String>,
+}
+
+impl LibraryCApiConfig {
+    pub fn sover(&self) -> anyhow::Result<String> {
+        let major = self.version.major;
+        let minor = self.version.minor;
+        let patch = self.version.patch;
+
+        let sover = match self.version_suffix_components {
+            None => match (major, minor, patch) {
+                (0, 0, patch) => format!("0.0.{patch}"),
+                (0, minor, _) => format!("0.{minor}"),
+                (major, _, _) => format!("{major}"),
+            },
+            Some(1) => format!("{major}"),
+            Some(2) => format!("{major}.{minor}"),
+            Some(3) => format!("{major}.{minor}.{patch}"),
+            Some(num) => anyhow::bail!("Unexpected number of suffix components: {num}"),
+        };
+
+        Ok(sover)
+    }
 }
 
 #[derive(Debug, Default)]
@@ -619,6 +643,7 @@ fn load_manifest_capi_config(pkg: &Package) -> anyhow::Result<CApiConfig> {
     let mut version = pkg.version().clone();
     let mut install_subdir = None;
     let mut versioning = true;
+    let mut version_suffix_components = None;
     let mut import_library = true;
     let mut rustflags = Vec::new();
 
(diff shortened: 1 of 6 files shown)


kyrias:
Haven't looked at the code, but the specification in the description sounds great.  :+1:

src/build.rs
```diff
@@ -406,10 +407,33 @@ pub struct LibraryCApiConfig {
...
+        let sover = match self.version_suffix_components {
+            None => match (major, minor, patch) {
+                (0, 0, patch) => format!("0.0.{patch}"),
+                (0, minor, _) => format!("0.{minor}"),
+                (major, _, _) => format!("{major}"),
+            },
+            Some(1) => format!("{major}"),
+            Some(2) => format!("{major}.{minor}"),
+            Some(3) => format!("{major}.{minor}.{patch}"),
+            Some(num) => anyhow::bail!("Unexpected number of suffix components: {num}"),
```
lu-zero, line 430: If you want to go the extra mile, you should put the validation when parsing the value instead.

lu-zero (commented):
Overall it looks good, there is only a nit I'd like to have addressed, if you do not have time I can do that myself later though.

Thank you a lot for the patch :)

src/build.rs
kpcyrd: Good idea, done!

lu-zero (approved):
Thank you a lot!

PR #367, opened 2024-02-29
Match the beaviour of meson regarding relative paths in the install directories

### LINKED ISSUE #366: cinstall does not append prefix to relative libdir
`cargo cinstall --prefix=/opt/foo --libdir=lib --release --library-type cdylib --destdir=/home/xclaesse/root`

When running that command in gst-plugins-rs, I'm expecting installed libraries to go into `/home/xclaesse/root/opt/foo/lib`. However it seems to put them in `/home/xclaesse/root/lib`. 

Appending the prefix to destdir is what `meson install` or autotools typically do. Note that it's far from easy to get that right on Windows, be careful.

lu-zero:
Fixes #366

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -16,27 +16,32 @@ struct Common {
     /// Path to directory where target should be copied to
     #[clap(long = "destdir")]
     destdir: Option<PathBuf>,
-    /// Directory path used to construct default values of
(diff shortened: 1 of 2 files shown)


sdroege:
Behaviour-wise this makes sense and is what I would expect. I've run into this problem before when I wrongly assumed relative paths for libdir etc work :)

PR #373, opened 2024-05-04
Implement Meson parity

### LINKED ISSUE #279: cinstall doesn't install PDB debug symbol files with MSVC
With MinGW, the debug information is embedded in the object files, but with MSVC it is split out into PDB files. These need to be installed alongside the object files to get symbols in the Visual Studio debugger. Some references to how meson handles this:

https://github.com/mesonbuild/meson/blob/a692395186b9a74817a584444418901da1d7306f/mesonbuild/backend/ninjabackend.py#L2427-L2474

https://github.com/mesonbuild/meson/blob/a692395186b9a74817a584444418901da1d7306f/mesonbuild/backend/ninjabackend.py#L3069-L3074

Compile args:

https://github.com/mesonbuild/meson/blob/a692395186b9a74817a584444418901da1d7306f/mesonbuild/compilers/mixins/visualstudio.py#L304-L308

https://github.com/mesonbuild/meson/blob/a692395186b9a74817a584444418901da1d7306f/mesonbuild/compilers/mixins/visualstudio.py#L410-L420

Linker args:

https://github.com/mesonbuild/meson/blob/a692395186b9a74817a584444418901da1d7306f/mesonbuild/linkers/linkers.py#L1225-L1230

tl;dr: if you install foo.dll you must also install foo.pdb next to it. Same with foo.exe. You can ignore static libraries.

### LINKED ISSUE #280: Consider naming DLLs built with MinGW as `libfoo.dll`

### LINKED ISSUE #317: cargo-c: improve install path conventions
When installing the `gst-plugin-rtp` plugin from https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs, I noticed that with:
```cargo cinstall -p gst-plugin-rtp --prefix=/usr/local```

the rust plugins (the shared libs) install to `/usr/local/lib/gstreamer-1.0/libgstrsrtp.so`

whereas other gstreamer packages install to e.g., `/usr/local/lib/x86_64-linux-gnu/gstreamer-1.0/`

For now I can workaround this by explicitly adding ` --libdir=${PREFIX}/lib/x86_64-linux-gnu` but it'd be preferable for this to follow the `meson` + `autotools` conventions (at least as I understand them).

@sdroege's comment:
> there's probably some useful heuristic to detect lib, lib32, lib64, lib/target-triple

### PR DESCRIPTION
**Pull request by amyspark**
Hi @lu-zero,

This PR tries to implement the necessary bits to support the following Meson conventions:

- autodetection of `libdir` under Debian distros
- installation of PDB symbols under MSVC
- calling static libraries `libfoo.a` and import libraries `foo.lib` under MSVC

The second item is done in a way that allows adding support for other OSes (as what's needed is passing `-C split-debuginfo=packed`, and this is the default on macOS).

Setting it as draft before testing; will do so later next week, but I wanted to get some early feedback first.

cc @nirbheek 

Fixes #279 
Fixes #280 
Fixes #317

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -30,7 +30,7 @@ struct Common {
     includedir: PathBuf,
     /// Path to directory for installing generated executable files
     #[clap(long = "bindir", default_value = "bin")]
-    bindir: PathBuf,
+    bindir: Option<PathBuf>,
     /// Path to directory for installing generated pkg-config .pc files
     ///
     /// [default: {libdir}/pkgconfig]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -9,6 +9,8 @@ use cargo_util::{ProcessBuilder, ProcessError};
 
 use clap::{Arg, ArgAction, ArgMatches, Command, CommandFactory, Parser};
 
+use crate::target::Target;
+
 // TODO: convert to a function using cargo opt()
 #[allow(dead_code)]
 #[derive(Clone, Debug, Parser)]
@@ -56,7 +58,8 @@ struct Common {
 }
 
 fn base_cli() -> Command {
-    Common::command()
+    let default_target = Target::new::<&str>(None, false);
+    let app = Common::command()
         .allow_external_subcommands(true)
         .arg(flag("version", "Print version info and exit").short('V'))
         .arg(flag("list", "List installed commands"))
@@ -107,7 +110,24 @@ fn base_cli() -> Command {
         .arg_target_dir()
         .arg_manifest_path()
         .arg_message_format()
-        .arg_build_plan()
+        .arg_build_plan();
+
+    if let Ok(t) = default_target {
+        app.mut_arg("prefix", |a| {
+            a.default_value(t.default_prefix().as_os_str().to_os_string())
+        })
+        .mut_arg("libdir", |a| {
+            a.default_value(t.default_libdir().as_os_str().to_os_string())
+        })
+        .mut_arg("datadir", |a| {
+            a.default_value(t.default_datadir().as_os_str().to_os_string())
+        })
+        .mut_arg("includedir", |a| {
+            a.default_value(t.default_includedir().as_os_str().to_os_string())
+        })
+    } else {
+        app
+    }
 }
 
 pub fn subcommand_build(name: &'static str, about: &'static str) -> Command {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -50,6 +50,9 @@ struct Common {
     #[clap(long = "crt-static")]
     /// Build the library embedding the C runtime
     crt_static: bool,
+    /// Use the Linux/Meson library naming convention on Windows
+    #[clap(long = "meson-paths", default_value = "false")]
+    meson: bool,
 }
 
 fn base_cli() -> Command {
--- a/src/install.rs
+++ b/src/install.rs
@@ -6,6 +6,7 @@ use cargo_util::paths::{copy, create_dir_all};
 
 use crate::build::*;
 use crate::build_targets::BuildTargets;
+use crate::target::Target;
 
 fn append_to_destdir(destdir: Option<&Path>, path: &Path) -> PathBuf {
     if let Some(destdir) = destdir {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/install.rs [hunk 1]
```diff
@@ -225,10 +228,43 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
 
         if let Some(ref static_lib) = build_targets.static_lib {
             ws.gctx().shell().status("Installing", "static library")?;
-            copy(
-                static_lib,
-                install_path_lib.join(static_lib.file_name().unwrap()),
-            )?;
+            let lib_type = LibType::from_build_targets(build_targets);
```
lu-zero, line 231: Probably this could live in `src/build_targets.rs`

src/install.rs
```diff
@@ -225,10 +228,43 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
...
+
+            copy(static_lib, install_path_lib.join(file_name))?;
+        }
+
+        if let Some(ref debug_info) = build_targets.debug_info {
+            ws.gctx()
+                .shell()
+                .status("Installing", "debugging information")?;
+            let lib_type = LibType::from_build_targets(build_targets);
+            match lib_type {
```
lu-zero, line 251: this logic also may live in `src/build_targets.rs`.

src/install.rs
```diff
@@ -241,7 +277,11 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
                     lib.install(capi_config, shared_lib, &install_path_lib)?;
                 }
                 LibType::Windows => {
-                    let lib_name = shared_lib.file_name().unwrap();
+                    let lib_name = if build_targets.use_meson_naming_convention {
```
lu-zero, line 280: Same here.

src/install.rs
```diff
@@ -281,15 +325,102 @@ pub struct InstallPaths {
 }
 
 impl InstallPaths {
-    pub fn new(_name: &str, args: &ArgMatches, capi_config: &CApiConfig) -> Self {
+    fn is_debianlike(target: &Target) -> bool {
+        consts::ARCH.eq_ignore_ascii_case(&target.arch)
```
lu-zero, line 329: to make it safer or less surprising that should trigger only if --target is not passed. (and needs to be documented)

src/install.rs
```diff
@@ -281,15 +325,102 @@ pub struct InstallPaths {
...
+
+    fn is_freebsd(target: &Target) -> bool {
+        target.os.eq_ignore_ascii_case("freebsd")
+    }
+
+    fn is_haiku(target: &Target) -> bool {
+        target.os.eq_ignore_ascii_case("haiku")
+    }
+
+    fn is_windows(target: &Target) -> bool {
```
lu-zero, line 342: In general those methods should be part of Target.

src/cli.rs
```diff
@@ -20,13 +20,13 @@ struct Common {
     /// `bindir`, `datarootdir`, `includedir`, `libdir`
     ///
     /// If they are absolute the prefix is ignored.
-    #[clap(long = "prefix", default_value = "/usr/local")]
+    #[clap(long = "prefix")]
```
lu-zero, line 23: That would require to use `default_value_if` to make it user friendly or to change the types to Option<T>.

A logic that could be simpler might be:
- Build a Target from [consts](https://doc.rust-lang.org/std/env/consts/index.html#)
- Produce the host defaults and the meson-defaults and set them all using `default_value_if`.
- If --target is present override accordingly.

src/install.rs
amyspark: *nods* I'll implement an approach in a separate commit. Let me know if it looks OK.

src/cli.rs
amyspark: I'm not exactly sure I follow, as your use of `clap::derive` here means that any changes must be doable at build time only. Can you explain it a bit more, once I submit the revised approach?

lu-zero: Sure, Parser imples CommandFactory so you can use [command](https://docs.rs/clap/latest/clap/trait.CommandFactory.html#tymethod.command) so you can still extend the cli

amyspark: Ah, then it's definitely not possible (AFAIU).

Currently you obtain Rust's host's information through cargo's `cargo::util::command_prelude::ArgMatchesExt.workspace`. That of course requires instantiating `Command::command()` itself, which is also the instance requiring the default value.

src/target.rs
```diff
@@ -113,4 +116,79 @@ impl Target {
...
+    fn is_haiku(&self) -> bool {
+        self.os.eq_ignore_ascii_case("haiku")
+    }
+
+    fn is_windows(&self) -> bool {
+        self.os.eq_ignore_ascii_case("windows")
+    }
+
+    pub fn default_libdir(&self) -> PathBuf {
+        if self.is_target_overridden {
```
lu-zero, line 139: It should be `if !is_target_overridden`, then do the debianlike process, run the rest of the checks in any case.

src/target.rs
```diff
@@ -134,24 +139,22 @@ impl Target {
     }
 
     pub fn default_libdir(&self) -> PathBuf {
+        if !self.is_target_overridden || self.is_freebsd() {
```
lu-zero, line 142: if the target is overridden I would disable the heuristic.

lu-zero:
I fixed most I could, but I do not have window to fix the last bit.

lu-zero:
Looks like there is some `_` vs `-` ongoing.

Issue #379, opened 2024-06-19
0.10.0 changes libdir unexpectedly


cpu:
Hi folks,

Over in rustls-ffi we have a [GitHub workflow](https://github.com/rustls/rustls-ffi/blob/5d445f97fd94279f6de41d5e78084ca5c4c9b393/.github/workflows/pkg-config.yaml) and [Makefile](https://github.com/rustls/rustls-ffi/blob/5d445f97fd94279f6de41d5e78084ca5c4c9b393/Makefile.pkg-config) that tests building a `.so`/`.pc` for rustls and then building our demo applications with it + `pkg-config`. 

I noticed that after the job switched to installing the just-released [0.10.0](https://github.com/lu-zero/cargo-c/releases/tag/v0.10.0) that it [started to fail](https://github.com/rustls/rustls-ffi/actions/runs/9582189667/job/26430694732) with output indicating `pkg-config` isn't finding the `.pc` file anymore.  [Switching the job to use 0.9.32](https://github.com/rustls/rustls-ffi/pull/435) fixed the build, so I'm fairly confident that something about rustls-ffi and cargo-c 0.10.0 is at fault. 

Confusingly, when I reproduce the build locally on my NixOS system with  cargo 1.81.0-nightly, pkg-config 0.29.2 and cargo-c 0.10.0 everything seems to "just work" (?).  The `.pc` file produced in my local testing seems identical between versions (with the exception of `libdir` changing from `libdir=${exec_prefix}/lib` (0.9.29) to `libdir=${prefix}/lib` (0.10.0) (but since `prefix == exec_prefix` in the `.pc` this change doesn't feel particularly meaningful).

Do you have any thoughts or ideas for next debugging steps? Thanks!

cpu:
I think it's perhaps specific to using `cargo-c` 0.10.0 on a Debian-like system, which would help explain why I can't reproduce locally on my non-Debian system. 

Adding more debugging to CI I see that `cargo cinstall --locked --release --prefix=/tmp/librustls` using 0.10.0 is emitting a directory structure like:

```
/tmp/librustls
/tmp/librustls/lib
/tmp/librustls/lib/x86_64-linux-gnu
/tmp/librustls/lib/x86_64-linux-gnu/librustls.so
/tmp/librustls/lib/x86_64-linux-gnu/librustls.so.0.13.0
/tmp/librustls/lib/x86_64-linux-gnu/pkgconfig
[...]

PR #380, opened 2024-06-19
Add a note to warn about the behaviour change on debian-like systems

lu-zero:
Addresses #379.


lu-zero:
@cpu Let me know if such note would had helped.

cpu (approved):
Thank you. I think this would have helped :+1:

PR #381, opened 2024-06-25
Commit Cargo.lock

mwcampbell:
I believe the current standard practice, especially for binary/executable projects, is to commit Cargo.lock. So this PR does that, and modifies .gitignore appropriately.


lu-zero:
It is a practice I disagree with since it causes too much churn because `Cargo.lock` is diff-unfriendly, and does not provide anything of value since we release frequently.

PR #382, opened 2024-06-26
Allow debug symbols install to be skipped if they are not present

amyspark:
Hi,

This PR is to address an issue found in https://github.com/AOMediaCodec/libavif/pull/2228#discussion_r1653512613 . PDB symbols being missing (due to the `release` profile or `strip` being set) should not cause an install failure.

Let me know what you think.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -270,15 +270,21 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
         }
 
         if let Some(ref debug_info) = build_targets.debug_info {
-            ws.gctx()
-                .shell()
-                .status("Installing", "debugging information")?;
-            let destination_path = build_targets
-                .debug_info_file_name(&install_path_bin, &install_path_lib)
-                .unwrap();
+            if debug_info.exists() {
+                ws.gctx()
+                    .shell()
+                    .status("Installing", "debugging information")?;
+                let destination_path = build_targets
+                    .debug_info_file_name(&install_path_bin, &install_path_lib)
+                    .unwrap();
 
-            create_dir_all(destination_path.parent().unwrap())?;
-            copy(debug_info, destination_path)?;
+                create_dir_all(destination_path.parent().unwrap())?;
+                copy(debug_info, destination_path)?;
+            } else {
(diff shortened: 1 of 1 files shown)


src/install.rs
```diff
@@ -270,15 +270,21 @@ pub fn cinstall(ws: &Workspace, packages: &[CPackage]) -> anyhow::Result<()> {
...
+                let destination_path = build_targets
+                    .debug_info_file_name(&install_path_bin, &install_path_lib)
+                    .unwrap();
+
+                create_dir_all(destination_path.parent().unwrap())?;
+                copy(debug_info, destination_path)?;
+            } else {
+                ws.gctx()
+                    .shell()
+                    .status("Skipping", "debugging information")?;
```
lu-zero, line 286:
```suggestion
                    .status("Absent", "debugging information")?;
```

Maybe?

lu-zero (approved):

lu-zero:
This weekend I will cut a 0.10.1 with this fix.

wantehchang:
Thank you for the fix. I have tested this commit in https://github.com/AOMediaCodec/libavif/pull/2228. Now the rav1e build succeeds with the following messages in the [output log](https://github.com/AOMediaCodec/libavif/actions/runs/9686263235/job/26728285271?pr=2228) (please expand "Build libavif (ninja)"):

```
  Installing pkg-config file
  Installing header file
  Installing static library
      Absent debugging information
```

Note that "Absent" is not aligned with "Installing". Is that indentation expected?

lu-zero:
Yes, the first word is a tag. probably it could be made a a debug message instead though..

wantehchang:
Did you mean the "Installing" tag is not indented, but the "Absent" tag is indented?

lu-zero:
The first word is a right-aligned tag that has a color (green), the rest is a left-aligned message.

https://docs.rs/cargo/0.80.0/cargo/core/shell/struct.Shell.html#method.status

wantehchang:
Thank you for the explanation. I didn't notice that 'Installing" and "Absent" are right-aligned. I only looked on the left.

lu-zero:
image

You should notice it with cargo itself as well :)

Issue #394, opened 2024-08-15
Panics when building a #![no_std] crate on linux


jahartsell:
It seems like some crates can have an empty link line, which causes a panic when determining the static libs here

https://github.com/lu-zero/cargo-c/blob/f873c83cfb637d38f25ef92fb5185aa309f7e111/src/build.rs#L1158

I assume static libs should just be empty in that case, and the following at least builds my case.

```diff
diff --git a/src/build.rs b/src/build.rs
index 273efc1..59b673c 100644
--- a/src/build.rs
+++ b/src/build.rs
@@ -1160,7 +1160,7 @@ pub fn cbuild(
                     .unwrap()
                     .values()
                     .next()
-                    .unwrap()
+                    .map_or("", |s| s)
                     .to_string()
             };
             let capi_config = &cpkg.capi_config;
```

The following example reproduces the issue for me using rust 1.80.1 and  cargo-c 0.10.3+cargo-0.81.0 on Ubuntu 22.04

Cargo.toml
```toml
[package]
name = "testlib"

[features]
capi = []

[package.metadata.capi.library]
rustflags = "-Cpanic=abort"
```

lib.rs
```rust
#![no_std]

#[cfg(feature="capi")]
mod capi {
    #[panic_handler]
    fn panic_handler(_info: &core::panic::PanicInfo) -> ! {
        loop {}
    }

    #[no_mangle]
    extern "C" fn example() -> u32 {
        0
    }
}
```

lu-zero:
If you can send a PR I'd land it before making the next release, sorry for the belated reply but sometimes github messages get lost in the mailbox :/

PR #396, opened 2024-09-04
Handle unwrap when link line is empty, i.e. nostd crates.

jahartsell:
A terribly tiny PR to follow up on [#394](https://github.com/lu-zero/cargo-c/issues/394)

This works for my no-std package, and the coverage script still passes. If there's anything else needed just let me know!

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -1158,7 +1158,7 @@ pub fn cbuild(
                     .unwrap()
                     .values()
                     .next()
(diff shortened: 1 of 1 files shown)


lu-zero:
I'd reword a little more the commit with at least Fixes #394, but I can do myself

lu-zero:
Thank you a lot :)

Issue #210, opened 2021-06-28
Make pkg-config path generation support cross compiling


lu-zero:
The paths are built using the os-specific `std::path::Path`, make pkg-config aware of the target and rewrite the paths using the correct separator for the target.

1480c1:
Compiling ffmpeg with mingw-w64-x86_64-clang from msys2 fails when compiling the rav1e libavcodec/librav1e.o due to not finding rav1e.h, mingw-w64-x86_64-gcc is fine.

The failure seems to stem from `-I${includedir}\rav1e` as changing that to `-I${includedir}/rav1e` makes the compilation successful. Unsure if the issue is that clang is somehow interpreting the `\r` as a carriage return or somewhere else.

PR #399, opened 2024-09-09
Cleanup pkgconfig output

folkertdev:


src/pkg_config_gen.rs
```diff
@@ -307,6 +282,102 @@ mod test {
...
+            let result = canonicalize(path);
+            assert_eq!(result, "/home/user/docs");
+        }
+
+        #[test]
+        fn test_parent_directory() {
+            let path = "/home/user/../docs";
+            let result = canonicalize(path);
+            assert_eq!(result, "/home/user/../docs");
+        }
```
folkertdev, line 332: is this expected? should that not output `/home/docs`?

src/pkg_config_gen.rs
```diff
@@ -5,27 +5,31 @@ use crate::install::InstallPaths;
...
+        match component {
             Component::RootDir => {
                 separator = "/";
-                String::new()
             }
-            Component::Prefix(_) => p.as_os_str().to_string_lossy().to_string(),
-            _ => {
-                let c = format!("{}{}", separator, p.as_os_str().to_string_lossy());
+            Component::Prefix(_) => {
+                write!(out, "{}", component_str).unwrap();
```
folkertdev, line 21: a write to a `String` will only fail when the system runs out of memory. The rust standard library generally disregards OOM, and it seems reasonable to do here too.

lu-zero: It shouldn't be a problem indeed.

src/pkg_config_gen.rs
lu-zero: It should, if you want to add more logic is great :)

lu-zero:
Looks like clippy is now unhappy, willing to add a patch before your set?

src/pkg_config_gen.rs
folkertdev: I'll play around with it a bit

folkertdev: got it to work. less pretty now but more functionality. I think this still gives weird results for e.g. `../` or any case where there are more parent directory components than there are normal components. should that panic maybe? or is that unlikely to come up?

lu-zero: Its origin is to canonicalize \-paths to /-paths so do not dwell on it your original set + the clippy fix would be already great.

src/pkg_config_gen.rs
```diff
@@ -307,6 +306,102 @@ mod test {
...
+            "Cflags: -I${includedir} -DFOO\n",
+            "Requires: somelib, someotherlib\n",
+            "Requires.private: someprivatelib >= 1.0\n",
+        );
+
+        assert_eq!(expected, pkg.render());
+    }
+
+    mod test_canonicalize {
+        use super::canonicalize;
```
lu-zero, line 328: It needs tests for "\a\b\c" and such paths.

lu-zero: But only on windows...

lu-zero (approved):
It is good enough as it is.

src/pkg_config_gen.rs
folkertdev: on windows, `a\b\c` gets parsed as a single normal component. Also pkgconfig is just for unix systems right, so windows paths don' t make a whole lot of sense even when cross-compiling. Maybe WSL causes complications?

lu-zero: pkgconfig exists and is used also on windows :) See https://github.com/lu-zero/cargo-c/issues/210 on why

PR #400, opened 2024-09-10
extract and test target file name creation

folkertdev:


src/build_targets.rs
```diff
@@ -69,82 +69,31 @@ impl BuildTargets {
...
-            }
-            _ => unimplemented!("The target {}-{} is not supported yet", os, env),
-        };
-
-        let static_lib = if libkinds.contains(&"staticlib") {
-            Some(static_lib)
-        } else {
-            None
+        let Some(file_names) = FileNames::from_target(target, name, targetdir) else {
+            unimplemented!(
```
lu-zero, line 78: This can be an `anyhow!` even if should be caught earlier already.

lu-zero (approved):
Overall looks good, if you want to go the extra mile please convert the unreachable to an error that can bubble up.

src/build_targets.rs
folkertdev: done

lu-zero (approved):

PR #401, opened 2024-09-10
path canonicalize windows tests

folkertdev:
my knowledge of windows paths is limited, and I used wine to develop these tests so that is a slightly different environment, but at least this shows that the basics work.


lu-zero (approved):
It looks fine, thank you!

PR #404, opened 2024-09-10
Generate import libraries without `Command`

folkertdev:
Gets rid of two more nasty `Command` invocations. This'll need a rebase once https://github.com/lu-zero/cargo-c/pull/402 is merged, but we can handle any review comments already. 

This time the logic is effectively untested by me, and would really benefit from the CI improvements suggested in the other PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ cc = "1.0"
 glob = "0.3"
 itertools = "0.13"
 object = "0.36.4"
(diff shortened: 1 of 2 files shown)


lu-zero (approved):
It looks fine to me beside the clippy lint.

lu-zero:
It will need a deprecation warning for the option but I can fold it in after myself.

PR #406, opened 2024-09-11
Fix pkg-config library output for MSVC and enable Windows CI

amyspark:
Hi @lu-zero ,

In https://github.com/lu-zero/cargo-c/pull/402#issuecomment-2343233329 I mentioned that the pkg-config generator needed a bit more work for it to work with MSVC.

This PR converts the output of `native-static-libs` to match the expected format of pkg-config (`-lfoo`, instead of `foo.lib`). I've also added the needed bit for this to be tested in CI.

Let me know what you think.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -16,6 +16,7 @@ use anyhow::Context as _;
 use cargo_util::paths::{copy, create_dir_all, open, read, read_bytes, write};
 use implib::def::ModuleDef;
 use implib::{Flavor, ImportLibrary, MachineType};
+use itertools::Itertools;
 use semver::Version;
 
 use crate::build_targets::BuildTargets;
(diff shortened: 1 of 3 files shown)


src/build.rs
```diff
@@ -1114,13 +1114,37 @@ pub fn cbuild(
...
+                        .to_string()
+                } else {
+                    exec.link_line
+                        .lock()
+                        .unwrap()
+                        .values()
+                        .next()
+                        .map_or("", |s| s)
+                        .to_string()
+                }
```
folkertdev, line 1147: I think it would be a good idea to move this logic into its own function

```suggestion
            let static_libs = if only_cdylib {
                String::new()
            } else if let Some(libs) = exec.link_line.lock().unwrap().values().next() {
                static_libraries(libs, &rustc_target)?
            } else {
                String::new()
            };
```

in combination with something like

```rust
fn static_libraries(
    link_line: &str,
    rustc_target: &target::Target,
) -> Result<String, std::fmt::Error> {
    use std::fmt::Write;

    let mut buf = String::with_capacity(256);

    for lib in link_line.trim().split(' ') {
        if lib.starts_with("/defaultlib") {
            continue;
        }

        if rustc_target.env == "msvc" && lib.ends_with(".lib") {
            write!(buf, "-l{} ", lib.trim_end_matches(".lib"))?;
        } else {
            write!(buf, "{lib} ")?;
        }
    }

    Ok(buf)
}
```

This function is then testable, so you can add some tests at the bottom of the module to check that this logic does the right thing, independent of the architecture that the code runs on.

amyspark: This is a good idea, I went with the functional approach because with it I can filter out duplicated and empty flags.

lu-zero (approved):

lu-zero:
The code looks good, why msvc is unhappy though?

amyspark:
@lu-zero the pkg-config module requires `-lfoo`, but the test isn't aware of this and blindly applies it to the linker. I need to re-convert it to `foo.lib`.

src/build.rs
```diff
@@ -1112,15 +1135,18 @@ pub fn cbuild(
...
-                    .map_or("", |s| s)
-                    .to_string()
+                String::new()
+            };
+            let capi_static_libs = if only_cdylib {
+                String::new()
+            } else if let Some(libs) = exec.link_line.lock().unwrap().values().next() {
+                libs.to_string()
+            } else {
+                String::new()
```
lu-zero, line 1149: Maybe this is more compact, but still readable:
```
let (pkg_config_static_libs, static_libs) = if ... { (String::new(), String::new()) } else if let ... { (static_libraries(libs), libs.to_string()) } else { ... }
```

lu-zero (approved):
Thank you a lot!

PR #409, opened 2024-09-11
add `LibraryTypes`

folkertdev:
why use vectors of strings when two booleans will do.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -103,7 +103,7 @@ fn build_pc_files(
 
 fn patch_target(
     pkg: &mut Package,
-    libkinds: &[&str],
+    library_types: LibraryTypes,
     capi_config: &CApiConfig,
 ) -> anyhow::Result<()> {
     use cargo::core::compiler::CrateType;
@@ -111,20 +111,19 @@ fn patch_target(
     let manifest = pkg.manifest_mut();
     let targets = manifest.targets_mut();
 
-    let kinds: Vec<_> = libkinds
-        .iter()
-        .map(|&kind| match kind {
-            "staticlib" => CrateType::Staticlib,
-            "cdylib" => CrateType::Cdylib,
-            _ => unreachable!(),
-        })
-        .collect();
+    let mut kinds = Vec::with_capacity(2);
 
-    for target in targets.iter_mut() {
-        if target.is_lib() {
-            target.set_kind(TargetKind::Lib(kinds.clone()));
-            target.set_name(&capi_config.library.name);
-        }
+    if library_types.staticlib {
+        kinds.push(CrateType::Staticlib);
+    }
+
+    if library_types.cdylib {
+        kinds.push(CrateType::Cdylib);
+    }
+
+    for target in targets.iter_mut().filter(|t| t.is_lib()) {
+        target.set_kind(TargetKind::Lib(kinds.to_vec()));
+        target.set_name(&capi_config.library.name);
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


src/build.rs
```diff
@@ -998,6 +997,54 @@ fn deprecation_warnings(ws: &Workspace, args: &ArgMatches) -> anyhow::Result<()>
...
+    ) -> Self {
+        let (mut staticlib, mut cdylib) = (false, false);
+
+        for library_type in library_types {
+            staticlib |= library_type.as_ref() == "staticlib";
+            cdylib |= library_type.as_ref() == "cdylib";
+        }
+
+        // in this case, a cdylib cannot be produced
+        cdylib &= target.os != "none";
```
folkertdev, line 1034: based on the original logic, there is no check here for "musl", but there is in `from_target`. Under what circumstances can musl build a cdylib?

src/build_targets.rs
```diff
@@ -66,6 +66,24 @@ impl BuildTargets {
...
+    ) -> anyhow::Result<BuildTargets> {
+        Self::new_help(
+            name,
+            target,
+            targetdir,
+            LibraryTypes::from_library_types(target, libkinds.iter()),
+            capi_config,
+            use_meson_naming_convention,
+        )
+    }
```
folkertdev, line 78: this duplication is to prevent changes to the public api (which would require major version bumps). If that is not a concern then that would save a bunch of lines here. Of course this can always be resolved at some point in the future.

src/build.rs
lu-zero: when musl is being used as main libc, e.g. in Alpine, Gentoo and OpenWRT :)

src/build_targets.rs
lu-zero: It is fairly unlikely that somebody uses cargo-c as library, I'd call it `new_inner` while at it.

src/build.rs
```diff
@@ -998,6 +997,54 @@ fn deprecation_warnings(ws: &Workspace, args: &ArgMatches) -> anyhow::Result<()>
...
+
+impl LibraryTypes {
+    fn from_target(target: &target::Target) -> Self {
+        Self {
+            staticlib: true,
+            cdylib: target.os != "none" && target.env != "musl",
+        }
+    }
+
+    fn from_config(target: &target::Target, args: &ArgMatches) -> Self {
```
lu-zero, line 1015: This should be named `from_args` maybe...

src/build_targets.rs
folkertdev: well if you don't care about bumping the major version then this intermediate layer can just be removed right?

lu-zero:
we shouldn't have additional library types to deal with so it should be fine to use two booleans, the string were inherited from how cargo dealt with that.

src/build_targets.rs
lu-zero: it isn't a concern IMHO, if somebody complains because it is using cargo-c as library I'll bump the version. The target users shouldn't have problems anyway.

src/build.rs
folkertdev: cool, I added some comments with this information

src/build.rs
```diff
@@ -998,6 +997,58 @@ fn deprecation_warnings(ws: &Workspace, args: &ArgMatches) -> anyhow::Result<()>
...
+    pub staticlib: bool,
+    pub cdylib: bool,
+}
+
+impl LibraryTypes {
+    fn from_target(target: &target::Target) -> Self {
+        // for os == "none", cdylib does not make sense. By default cdylib is also not built on
+        // musl, but that can be overriden by the user. That is useful when musl is being used as
+        // main libc, e.g. in Alpine, Gentoo and OpenWRT
+        Self {
```
lu-zero, line 1012: You may link https://github.com/lu-zero/cargo-c/issues/180 or the README section, I wish there is a better way to see if the rustc in use defaults one way or the other w/out trying beforehand :/

lu-zero (approved):
Thank you!

PR #410, opened 2024-09-12
reduce the number of feature flags enabled for `object`

folkertdev:
also sets the version to the version used by `implib`, so `object` does not get built twice

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ anyhow = "1.0"
 cc = "1.0"
 glob = "0.3"
 itertools = "0.13"
-object = "0.36.4"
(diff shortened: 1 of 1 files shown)


lu-zero:
I hope implib gets a new release soon.

folkertdev:
why? why would any of its functionality break or need updating? these formats haven't changed in ages.

lu-zero:
I sent a pr to use the current `object`

folkertdev:
I've now updated this PR with the latest versions (so the `implib` release from yesterday with the newer `object` version)

lu-zero:
Perfect :)

lu-zero (approved):

PR #413, opened 2024-10-14
pkg-config: Use the right link-line when generating pkg-config files

thiblahute:
We were always using the first package link line, leading to broken pkg-config files.

For example in
https://gitlab.freedesktop.org/thiblahute/gst-plugins-rs/-/jobs/64909911, we can see that static linking using that `.pc` file doesn't work as it is not linking to `stdc++` in that specific case.


lu-zero:
Thank you for spotting and fixing it!

thiblahute:
Thanks for merging!

ooc, when do you think that should be released?

lu-zero:
Either when the new cargo is released or sooner if you really need it.

PR #416, opened 2024-10-17
Bump cargo version

lu-zero:


sdroege:
cargo 0.83 requires Rust 1.80, so you need to update the MSRV in Cargo.toml too

lu-zero:
Thank you for reminding me!

PR #423, opened 2024-11-03
Clarify that the `TL;DR` section is not a list of requirements

### LINKED ISSUE #422: Explanation for configuration?
So, I see that the [README][r] gives a tl;dr of what needs to be done to use cargo-c.

Compared to how simple it is to get started with cbindgen, there seem to be a lot of steps (and dependencies) required, and I wonder if it might be worth explaining to end-users why cargo-c is configured this way.

To generate headers with cbindgen requires very little ceremony at all. As long as you can get the executable from somewhere, you can just invoke it from within a rust project, and it will output a header file. You don't actually need to write a `cbindgen.toml` file (though it's often handy to do so), and you don't need to alter your project in any way if you don't want to (other than making sure there are at least some `#[no_mangle] pub` functions or constants).

So why does cargo-c require that we create a `capi.rs` – why can't it just use the same functions etc. that cbindgen uses? Why does cargo-c require adding `#[cfg(feature="capi")]` to the project? Why does it require\* creation of a `cbindgen.toml` file, when cbdingden itself doesn't?

\* Or maybe it's just a suggestion, the README isn't clear. It says "Remember to add a cbindgen.toml", but doesn't make clear whether this is a requirement or only useful to do.

I may also have a misunderstanding of what cargo-c is doing, but it also seems odd to me that it should require OpenSSL headers or `git` as prerequisites. If its job is to "produce and install a correct [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) file, a static library and a dynamic library, and a C header" – the main addition over what cargo and cbindgen can already do seems to be generation of a `.pc` file, and that's just a text generation task, isn't it? Why should cargo-c need access to git or OpenSSL for that? No doubt there are good reasons, but it might be helpful to make them clear. A simple `.pc` generator [like this one](https://github.com/nmante/pkg-config-generator) doesn't require access to git tooling or OpenSSL headers, so why does cargo-c?

[r]: https://github.com/lu-zero/cargo-c/blob/d30f40096565acb7ab77e304a9f8a2ce7c8dbc31/README.md

lu-zero:
Fixes #422.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For a more in-depth explanation of how `cargo-c` works and how to use it for
 your crates, read [Building Crates so they Look Like C ABI Libraries][dev.to].
 
 ### The TL;DR:
-
+This is the ideal setup for a project that wants to keep their C-API within the main crate:
 - [Create][diff-1] a `capi.rs` with the C-API you want to expose and use
   ~~`#[cfg(cargo_c)]`~~`#[cfg(feature="capi")]` to hide it when you build a normal rust library.
 - [Make sure][diff-2] you have a lib target and if you are using a workspace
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,10 +65,16 @@ your crates, read [Building Crates so they Look Like C ABI Libraries][dev.to].
   to install `cargo-c` and do `cargo cinstall --prefix=/usr
   --destdir=/tmp/some-place` or something along those lines.
 
+If you plan to keep the bindings as a separate crate and do not need to autogenerate the headers you may just [populate Cargo.toml][diff-5]:
```
sdroege, line 68: If you need an example of same-crate-but-not-autogenerated-header then you could link to ebur128

sdroege (approved):

PR #427, opened 2024-11-29
Add the example workspace to the CI

lu-zero:


lu-zero:
It isn't needed in the end.

PR #428, opened 2024-11-29
Unbreak .def generation

lu-zero:


lu-zero:
@cpu could you please try this and report back?

cpu:
@lu-zero I built/installed this branch and it worked great for my example-ffi test repo :tada: Thanks for the very quick fix.

cpu:
>  it worked great for my example-ffi test repo

And for what it's worth, also for `rustls-ffi`, where I started down this road originally 🔒

lu-zero:
Good to know, in case other problems arise:
- https://github.com/mstorsjo/msvc-wine is a boon
- https://github.com/rust-cross/cargo-xwin is also quite interesting to setup a msvc toolchain
- we could copy more mingw-w64 `gendef`

cpu:
>  in case other problems arise:

Thanks for sharing these resources :bow:

Issue #426, opened 2024-11-28
cargo-c produces a corrupt dll.lib on Windows?


cpu:
Hi folks,

_Please bear with me: I'm very inexperienced with both Windows and `cmake` and could be doing something laughably wrong :-)_

I'm trying to build a Rust library as a `.dll` using `cargo capi install`. My goal is to use that DLL from a C application built with MSVC and CMake that dynamically links the `.dll`. In a small [example repo I made](https://github.com/cpu/example-ffi) this results in an error of the form "example-ffi.dll.lib : fatal error LNK1127: library is corrupt". 

## Versions 
OS:  Microsoft Windows 11 Home (10.0.22631 Build 22631)
Arch: x86_64

```
cargo version
cargo 1.83.0 (5ffbef321 2024-10-29)
```
```
cargo capi install --version
cargo-c 0.10.5+cargo-0.83.0
```

```
cmake --version
cmake version 3.30.5
```

```
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.22631.
-- The C compiler identification is MSVC 19.41.34123.0
```

## Reproduction steps

I've outlined the steps in https://github.com/cpu/example-ffi but they basically boil down to:

* Using `cargo capi install` to spit out the expected `lib`, `bin` and `include` directories.
* Using `cmake` [find_library](https://cmake.org/cmake/help/latest/command/find_library.html) to locate the library
* Using `cmake` [target_link_library](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) to link the library

At this point an error like the following is produced:

```
MSBuild version 17.11.9+a69bbaaf5 for .NET Framework

  1>Checking Build System
  Building Custom Rule C:/Users/danie/Code/Rust/example-ffi/example-c/CMakeLists.txt
  main.c
C:\Users\danie\Code\Rust\example-ffi\example-ffi\build\lib\example-ffi.dll.lib : fatal error LNK1127: library is corrupt [C:\Us
ers\danie\Code\Rust\example-ffi\example-c\build\example-c.vcxproj]
```

Trying to look at the generated `.dll.lib` file using `dumpbin /exports example-ffi.dll.lib` produces the same error, hinting that it's perhaps not a `cmake` issue, but something else:

```
Microsoft (R) COFF/PE Dumper Version 14.41.34123.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file example-ffi.dll.lib

[...]

PR #432, opened 2025-01-07
Fix Meson naming convention

kleisauke:
This PR fixes two issues when using `--meson-paths`.

- Ensure import libraries under MinGW are named `libfoo.dll.a` (44e96dcd46a6e2385e328d1bc914707b4c04a8c9).
  ```diff
  @@ -4,5 +4,5 @@
        Copying /tmp/rav1e/x86_64-pc-windows-gnullvm/release/rav1e.h to /usr/x86_64-w64-mingw32/include/rav1e/rav1e.h
     Installing shared library
        Copying /tmp/rav1e/x86_64-pc-windows-gnullvm/release/rav1e.dll to /usr/x86_64-w64-mingw32/bin/librav1e.dll
  -     Copying /tmp/rav1e/x86_64-pc-windows-gnullvm/release/rav1e.dll.a to /usr/x86_64-w64-mingw32/lib/rav1e.lib
  +     Copying /tmp/rav1e/x86_64-pc-windows-gnullvm/release/librav1e.dll.a to /usr/x86_64-w64-mingw32/lib/librav1e.dll.a
        Copying /tmp/rav1e/x86_64-pc-windows-gnullvm/release/rav1e.def to /usr/x86_64-w64-mingw32/lib/rav1e.def
  ```
- Shared libraries under MSVC are always in the form `foo.dll` (9cf2de8126d9705cb78b3545ebed84d0e5c625d8).
  ```diff
  @@ -3,7 +3,7 @@
     Installing header file
        Copying C:\Users\kleisauke\rav1e\target\x86_64-pc-windows-msvc\release\rav1e.h to C:\include\rav1e\rav1e.h
     Installing shared library
  -     Copying C:\Users\kleisauke\rav1e\target\x86_64-pc-windows-msvc\release\rav1e.dll to C:\bin\librav1e.dll
  +     Copying C:\Users\kleisauke\rav1e\target\x86_64-pc-windows-msvc\release\rav1e.dll to C:\bin\rav1e.dll
        Copying C:\Users\kleisauke\rav1e\target\x86_64-pc-windows-msvc\release\rav1e.lib to C:\lib\rav1e.lib
        Copying C:\Users\kleisauke\rav1e\target\x86_64-pc-windows-msvc\release\rav1e.def to C:\lib\rav1e.def
     Installing debugging information
  ```

/cc @amyspark

As an aside, I've encountered the same issue as #426 with the `x86_64-pc-windows-gnullvm` target. It looks like `-Wl,--output-def` here:
https://github.com/lu-zero/cargo-c/blob/c4e6c08da0c9a12d16b1e00666e5ce67e3d013a9/src/target.rs#L112-L118
also generates a `.def` file with a missing library name, at least on LLVM (I haven't checked Binutils):
https://github.com/llvm/llvm-project/blob/llvmorg-19.1.6/lld/COFF/MinGW.cpp#L173-L192

Perhaps we could somehow allow `implib` to override the import library name?:
https://github.com/messense/implib-rs/blob/v0.3.3/src/lib.rs#L109-L112

### DIFF
--- a/src/build_targets.rs
+++ b/src/build_targets.rs
@@ -129,10 +129,18 @@ impl BuildTargets {
     }
 
     pub fn shared_output_file_name(&self) -> Option<OsString> {
-        if self.shared_lib.is_some() && self.use_meson_naming_convention {
-            Some(format!("lib{}.dll", self.name).into())
-        } else {
-            Some(self.shared_lib.as_ref()?.file_name().unwrap().to_owned())
+        match self.lib_type() {
+            LibType::Windows => {
+                if self.shared_lib.is_some()
+                    && self.use_meson_naming_convention
+                    && self.target.env == "gnu"
+                {
+                    Some(format!("lib{}.dll", self.name).into())
+                } else {
+                    Some(self.shared_lib.as_ref()?.file_name()?.to_owned())
+                }
+            }
+            _ => Some(self.shared_lib.as_ref()?.file_name()?.to_owned()),
         }
     }
 }
--- a/src/build_targets.rs
+++ b/src/build_targets.rs
@@ -74,7 +74,9 @@ impl BuildTargets {
             None
         };
 
-        let Some(file_names) = FileNames::from_target(target, name, targetdir) else {
+        let Some(file_names) =
+            FileNames::from_target(target, name, targetdir, use_meson_naming_convention)
+        else {
             return Err(anyhow::anyhow!(
                 "The target {}-{} is not supported yet",
                 target.os,
@@ -145,7 +147,12 @@ struct FileNames {
 }
 
 impl FileNames {
-    fn from_target(target: &Target, lib_name: &str, targetdir: &Path) -> Option<Self> {
+    fn from_target(
+        target: &Target,
+        lib_name: &str,
+        targetdir: &Path,
+        use_meson_naming_convention: bool,
+    ) -> Option<Self> {
         let (shared_lib, static_lib, impl_lib, debug_info, def) = match target.os.as_str() {
             "none" | "linux" | "freebsd" | "dragonfly" | "netbsd" | "android" | "haiku"
             | "illumos" | "openbsd" | "emscripten" | "hurd" => {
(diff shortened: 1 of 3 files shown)


src/build_targets.rs
```diff
@@ -164,13 +179,19 @@ impl FileNames {
 
                 if target.env == "msvc" {
                     let static_lib = targetdir.join(format!("{lib_name}.lib"));
-                    let impl_lib = targetdir.join(format!("{lib_name}.dll.lib"));
+                    let impl_lib = match use_meson_naming_convention {
+                        false => targetdir.join(format!("{lib_name}.dll.lib")),
+                        true => targetdir.join(format!("{lib_name}.lib")),
+                    };
```
amyspark, line 185: This bit is not entirely correct, under the Meson convention import libs are `lib` but also static libs are changed to `a`.  https://mesonbuild.com/FAQ.html#why-does-building-my-project-with-msvc-output-static-libraries-called-libfooa

src/build_targets.rs
```diff
@@ -164,13 +179,19 @@ impl FileNames {
...
                     let pdb = Some(targetdir.join(format!("{lib_name}.pdb")));
 
                     (shared_lib, static_lib, Some(impl_lib), pdb, Some(def))
                 } else {
                     let static_lib = targetdir.join(format!("lib{lib_name}.a"));
-                    let impl_lib = targetdir.join(format!("{lib_name}.dll.a"));
+                    let impl_lib = match use_meson_naming_convention {
+                        false => targetdir.join(format!("{lib_name}.dll.a")),
+                        true => targetdir.join(format!("lib{lib_name}.dll.a")),
```
amyspark, line 193: Since it's a boolean, it'd be clearer to write something like this:

```suggestion
                    let impl_lib = if use_meson_naming_convention {
                        targetdir.join(format!("lib{lib_name}.dll.a"))
                     } else {
                        targetdir.join(format!("{lib_name}.dll.a"))
```

src/build_targets.rs
kleisauke: Static libs behavior is controlled by: 
https://github.com/lu-zero/cargo-c/blob/5a92c3f567a3b9eebb547d8fedcc7de09b042821/src/build_targets.rs#L118-L129

The `shared_lib`, `static_lib` or `pdb` declarations cannot be modified here, as they also define which library is copied from Rust. Consequently, this logic must always align with:
https://github.com/rust-lang/rust/blob/1.84.0/compiler/rustc_target/src/spec/base/windows_msvc.rs#L11-L15

src/build_targets.rs
kleisauke: Thanks! Incorporated with commit 5a92c3f567a3b9eebb547d8fedcc7de09b042821.

kleisauke:
PR #433 will fix the CI failures.

> As an aside, I've encountered the same issue as #426 with the `x86_64-pc-windows-gnullvm` target. It looks like `-Wl,--output-def` here:
> 
> https://github.com/lu-zero/cargo-c/blob/c4e6c08da0c9a12d16b1e00666e5ce67e3d013a9/src/target.rs#L112-L118
> 
> 
> also generates a `.def` file with a missing library name, at least on LLVM (I haven't checked Binutils):
> https://github.com/llvm/llvm-project/blob/llvmorg-19.1.6/lld/COFF/MinGW.cpp#L173-L192
> Perhaps we could somehow allow `implib` to override the import library name?: https://github.com/messense/implib-rs/blob/v0.3.3/src/lib.rs#L109-L112

Work-in-progress commit https://github.com/kleisauke/cargo-c/commit/6ad3303b0c3dab1d5ff02cb79ce76f75be32e9d3 seems to fix this.

lu-zero:
If you have time during the weekend I'd fold your fixes in the next release.

lu-zero:
Thank you I defer to @amyspark for the review since I do not know enough of windows to be sure :)

amyspark (approved):

PR #433, opened 2025-01-09
Unbreak the CI

kleisauke:
The [`msvc` branch](https://github.com/amyspark/pkgconf/branches) has been removed. Instead, let's use pkgconf from MSYS2 with contains a patch that also fixes that linked issue.


lu-zero:
Thank you

PR #437, opened 2025-01-11
feat: add `Cargo.lock` for reproducible builds and revert `cc` crate to 1.2.7 to fix osx build issue

chenrui333:
- add `Cargo.lock` for reproducible builds (as the project has binary artifacts, it does make sense also checkin the lockfiles for reproducible builds)
- also revert `cc` crate to 1.2.7 (seeing some macos sdk detection issue with 1.2.8 release on the homebrew side)

----

cc crate 1.2.8 error log ref

```
  warning: libnghttp2-sys@0.1.10+1.61.0: ToolExecError: Command env -u IPHONEOS_DEPLOYMENT_TARGET LC_ALL="C" "clang" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "--target=arm64-apple-macosx15.2" "-I" "nghttp2/lib/includes" "-I" "/private/tmp/cargo-c-20250111-7579-cqlrm/cargo-c-0.10.8/target/release/build/libnghttp2-sys-30549b8c9f7bcb4d/out/i/include" "-DNGHTTP2_STATICLIB" "-DHAVE_NETINET_IN" "-DHAVE_TIME_H" "-DHAVE_ARPA_INET_H" "-o" "/private/tmp/cargo-c-20250111-7579-cqlrm/cargo-c-0.10.8/target/release/build/libnghttp2-sys-30549b8c9f7bcb4d/out/i/lib/64375acaab57ea54-sfparse.o" "-c" "--" "nghttp2/lib/sfparse.c" with args clang did not execute successfully (status code exit status: 1).cargo:warning=clang: error: no such file or directory: '-isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk'
```

relates to:
- https://github.com/Homebrew/homebrew-core/pull/203932
- https://github.com/rust-lang/cc-rs/issues/1362


lu-zero:
I consider having Cargo.lock in the tree more problems than its advantages, if we have to pin cc I'd do it otherwise.

lu-zero:
I'd just bump to ensure we use the version with this https://github.com/rust-lang/cc-rs/pull/1364 in

chenrui333:
> I consider having Cargo.lock in the tree more problems than its advantages

what are the problems?

chenrui333:
> I'd just bump to ensure we use the version with this [rust-lang/cc-rs#1364](https://github.com/rust-lang/cc-rs/pull/1364) in

yeah, if there is lockfile, then it would be tricky to pin or downgrade crates like cc.

lu-zero:
> > I consider having Cargo.lock in the tree more problems than its advantages
> 
> what are the problems?

In general it cause additional churn for us upstream for next to no downstream gain since people should either use the source distribution from crates.io and/or, if upstream is slow at releasing, they still have to maintain their own lockfile.

lu-zero:
the updated cc is up now :)

PR #438, opened 2025-01-12
Fix import lib generation under MinGW

kleisauke:
Override `import_name` after parsing the `.def` file, as `-Wl,--output-def,` in LLVM and Binutils does not include a library name in the generated `.def` file.
https://github.com/lu-zero/cargo-c/blob/c37aec4d53076c3198c38e3d2cf187e35e199d95/src/target.rs#L112-L118
https://github.com/llvm/llvm-project/blob/llvmorg-19.1.6/lld/COFF/MinGW.cpp#L173-L192
https://github.com/bminor/binutils-gdb/blob/binutils-2_43_1/binutils/dlltool.c#L1813-L1872

Also bump `implib` while we are here, to ensure the import library works when targeting `i686-pc-windows-gnu{,llvm}`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ anyhow = "1.0"
 cc = "1.0"
 glob = "0.3"
 itertools = "0.14"
-implib = "0.3.3"
+implib = "0.3.4"
 object = { version = "0.36.4", default-features = false, features = ["std", "read_core", "pe"] }
 
 [features]
--- a/src/build.rs
+++ b/src/build.rs
@@ -147,20 +147,15 @@ fn build_def_file(
         // Create the .def output file
         let def_file = cargo_util::paths::create(targetdir.join(format!("{name}.def")))?;
 
-        write_def_file(name, dll_file, def_file)?;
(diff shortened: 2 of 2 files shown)


amyspark (approved):

lazka:
A bit of a long shot, but we are getting crashes with the new import libs since updating to 0.10.9, see https://github.com/msys2/MINGW-packages/issues/23358. I tried reverting this PR, but then it fails at dynamic link time (do I need to downgrade rust too?)

lu-zero:
If you could bisect the releases to find the last one working would be great, even better if you can can make a smaller reproducer.

lazka:
I just went back all versions, the last working version was 0.10.4, then linking is broken in general until 0.10.9, but then things crash at runtime.

lazka:
Since the working importlib is much simpler than the broken one, my guess is that https://github.com/lu-zero/cargo-c/pull/404 could be related

lu-zero:
Can you please compare the two outputs and tell me what changed?

lazka:
I've created a (smaller) reproducer: https://github.com/lazka/cargo-c-crash

PR #444, opened 2025-02-13
Bump implib

### LINKED ISSUE #443: mingw: newer importlibs lead to runtime crashes with newer cargo-c
(Initially reported here, creating an issue to summarize: https://github.com/lu-zero/cargo-c/pull/438#issuecomment-2652209591)

The problematic project is https://github.com/ImageOptim/libimagequant, a small reproducer is provided here: https://github.com/lazka/cargo-c-crash

Using the generated importlib to link to libimagequant results in a crash if `liq_version()` is called (~~maybe libimagequant is at fault here and the cargo-c change is just exposing a bug? also an option~~ -> other projects also fail the same way)

* cargo-c 0.10.4 creates a working import lib
* cargo-c 0.10.9 creates a broken import lib

Using cargo-c 0.10.9 and re-creating the importlib with dlltool (either from binutils or llvm) fixes the issue.

Here are the different import libs introspected with "llvm-objdump -x":

* working llvm importlib cargo-c 0.10.3: https://gist.github.com/lazka/89efb6328ac4208a4fa0c67b00cfab0e
* working gnu importlib cargo-c 0.10.3: https://gist.github.com/lazka/54d6225d1f14a1be5593ab1e1b265f81
* broken importlib cargo-c 0.10.9: https://gist.github.com/lazka/b44f77137d90304c57eafa11a8cc7fd9

lu-zero:
Fixes #443.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ anyhow = "1.0"
 cc = "1.0"
 glob = "0.3"
 itertools = "0.14"
-implib = "0.3.4"
+implib = "0.3.5"
 object = { version = "0.36.4", default-features = false, features = ["std", "read_core", "pe"] }
 
 [features]


lazka:
Thanks. I've backported it now, so no rush :)

PR #449, opened 2025-02-21
Bump cargo version

lu-zero:


lu-zero:
@amyspark can you figure out what broke for msvc ^^; ?

sdroege:
AFAIU it's missing linking to ws2_32.dll or how it was called

amyspark:
> AFAIU it's missing linking to ws2_32.dll or how it was called

Yes, this should've been exposed by the `native-static-libs` of rustc, I'll need to switch into Windows and get 1.85 to verify.

amyspark:
@lu-zero merging this beforehand should take care of it: https://github.com/lu-zero/cargo-c/pull/450

The issue was that `pkg-config` takes precedence over `pkgconf` in Meson, and for some reason the recent install also began pulling the former.

lu-zero:
the problem seems to persist :/

amyspark:
Ah, you're bumping the cargo version! I hadn't noticed that on top of `pkg-config` causing its own errors. One moment...

amyspark:
This result is from cargo 0.86 and it's plainly wrong:

imagen

lu-zero:
can you check what rustc is emitting?

amyspark:
I'm doing that, and it's weird:

- If I run `cargo cbuild`, the correct information is emitted both to the `-uninstalled` and final modules
- But if I run `cargo cinstall` afterwards, the information is overridden with emptiness
- Running `cargo cinstall` results in emitting the correct information directly

Investigating now what's the difference between (1) and (2).

lu-zero:
Can you confirm that before the cargo update the problem appears?

amyspark:
@lu-zero Regression here: https://github.com/lu-zero/cargo-c/pull/446 , the second try doesn't execute `rustc` (because it's all already built and cached) so no `native-static-libs` makes it to `cbuild`. That empties the `Libs.private` field.

lu-zero:
Thank you for finding the problem now it will be fun to do the proper cache invalidation...

Issue #458, opened 2025-04-30
Can't build cdylib on Alpine with musl


nirbheek:
Fwict, ever since https://github.com/lu-zero/cargo-c/pull/409, it's not possible to build cdylibs on Alpine with musl at all, even with `RUSTFLAGS="-C target-feature=-crt-static"`. Only static libs are created:

```rust
cdylib: target.os != "none" && target.env != "musl"
```

This contradicts the troubleshooting section of the README and the review comment on that PR: https://github.com/lu-zero/cargo-c/pull/409#discussion_r1755468680

lu-zero:
If you have time to send a PR with the fix I'll land it today.

nirbheek:
Not sure what the fix should be, afaict the only thing we can do is to make that line: `cdylib: target.os != "none"` and let users figure out the rest by reading the docs.

lu-zero:
That would be already a step in the right direction. Ideally we could probe rustc for `  "crt-static-default": true,` but it is nightly only right now.

PR #459, opened 2025-05-01
Always allow building cdylibs on musl

nirbheek:
Building cdylibs can only work on musl when the CRT is not statically linked: `RUSTFLAGS="-C target-feature=-crt-static"` however we have no way to detect that case.

It's better to always allow it, and let the user find the "troubleshooting" section of cargo-c when they encounter an error.

In nightly rust, we can detect this case and we can add it once it's in stable.

Fixes https://github.com/lu-zero/cargo-c/issues/458


lu-zero:
Thank you!

Issue #455, opened 2025-04-03
Making deterministic .pc modules


amyspark:
Hi,

Coming from https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/599#note_2838913 -- 

Currently, the .pc generation in cargo-c is as simple as capturing the linker line for a given crate, its `Requires` and `Requires.private` entries, and string interpolating the two together into the file. While this certainly works, it does not allow for reproducibility:

- the order of libraries in the linker line is unpredictable (depends on the Cargo parallelization AFAIU)
- the `Libs` and `Cflags` entries are never pruned for flags provided in `Requires` etc.
- libraries provided by crates e.g. windows-targets (which does not install its .lib) or libsodium-sys (which does not provide a .pc file) break static usage of the .pc module

My proposal in https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/2162 was to handle it on our end by manually parsing and deduplicating entries, however it was determined that it was best to handle this at the source.

@lu-zero could this be handled in `cargo-c` in a reasonable way?

sdroege:
In the worst case you could do the same in cargo-c that were doing in the gst-plugins-rs MR. I'm not sure there is a better way.

lu-zero:
I need more information on why it happens. Probably we have to sort the link line fragments in cargo first to solve the first problem, the other two seem very outside of what we can solve within cargo or cargo-c though.

amyspark:
> I need more information on why it happens

From the looks of it, it's in rustc where it happens. I'm trying to track down the generation of `native-static-libs` to figure out why.

amyspark:
So it seems that the code in `compiler\rustc_codegen_ssa\src\back\link.rs` does not control for duplicates at all when reducing the linker line. Something like this should work, given that at this stage the compiler is aware of each flag and hasn't yet joined them:

```diff
diff --git a/compiler/rustc_codegen_ssa/src/back/link.rs b/compiler/rustc_codegen_ssa/src/back/link.rs
index e2081ad7563..08cd2962ffc 100644
--- a/compiler/rustc_codegen_ssa/src/back/link.rs
+++ b/compiler/rustc_codegen_ssa/src/back/link.rs
@@ -12,7 +12,7 @@
 use regex::Regex;
 use rustc_arena::TypedArena;
 use rustc_ast::CRATE_NODE_ID;
[...]

PR #470, opened 2025-06-14
Implement pkg-config flag deduplication

amyspark:
Should reduce a lot of the noise in Cflags and Libraries.

See #455

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ glob = "0.3"
 itertools = "0.14"
 implib = "0.3.5"
 object = { version = "0.36.4", default-features = false, features = ["std", "read_core", "pe"] }
+pkg-config = "0.3.32"
 
 [features]
 default = []
--- a/src/build.rs
+++ b/src/build.rs
@@ -235,14 +235,14 @@ struct FingerPrint {
     root_output: PathBuf,
     build_targets: BuildTargets,
     install_paths: InstallPaths,
-    static_libs: String,
+    static_libs: Vec<String>,
     hasher: DefaultHasher,
 }
 
 #[derive(serde::Serialize, serde::Deserialize)]
 struct Cache {
     hash: String,
-    static_libs: String,
+    static_libs: Vec<String>,
 }
 
 impl FingerPrint {
@@ -262,7 +262,7 @@ impl FingerPrint {
             root_output: root_output.to_owned(),
             build_targets: build_targets.clone(),
             install_paths: install_paths.clone(),
-            static_libs: String::new(),
+            static_libs: vec![],
             hasher,
         }
     }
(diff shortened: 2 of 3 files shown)


lu-zero:
if it doesn't explode horribly once you have the usual permutation of `-Wl,-z,relro -Wl,-z,now -Wl,--as-needed`...

amyspark:
> if it doesn't explode horribly once you have the usual permutation of -Wl,-z,relro -Wl,-z,now -Wl,--as-needed...

The only corner case I have so far (tested with Windows, macOS ongoing) is that the `self.libs` entry coming from the cache fingerprint isn't split like pkg-config's, so I need to sort it out.

That one case of relro etc. sounds prime for testing with Android, I'll do that later today.

nirbheek:
Speaking as the guy who implemented Meson's de-dup of link args and cflags, this is very easy to get wrong. Did you use Meson's de-dup class as a reference when building this?

lu-zero:
@nirbheek do you have a test suite we can share?

nirbheek:
Meson's test case won't exactly map here, because reordering link args needs to be done carefully. You can have undefined-symbol errors if you get the order wrong, or pick up the library from the wrong prefix. Meson does these things to alleviate this risk:

1. We wrap library args with `-Wl,--start-group ... -Wl,--end-group`
2. We preserve the order of `-L -l` args when de-duping
3. We resolve `-L -l` sets from each `.pc` file immediately to an on-disk file (static or shared, depending on config)

(1) (2) can be done by cargo-c, but not (3) since cargo-c doesn't know which one to pick.

Anyway, the implementation is here: https://github.com/mesonbuild/meson/blob/master/mesonbuild/arglist.py

Cflags-related tests are here: https://github.com/mesonbuild/meson/blob/master/unittests/internaltests.py#L114-L279

Some of the library find/arg tests are here: https://github.com/mesonbuild/meson/blob/master/unittests/internaltests.py#L537-L675. The others are integration tests, not unit tests, so not useful for cargo-c.

lu-zero:
thank you a lot :)

amyspark:
> Did you use Meson's de-dup class as a reference when building this?

Not at all, I was not thinking of any downstream build system in particular when doing this. I relied on pkg_config's own resolution logic for this, and made sure that flags did not lose their semantics until concatenation was needed e.g. `-Wl,--start-group ... -Wl,--end-group` is meant to stay as a single piece of link flag.

The big blocker for doing a Meson-Rust port here is that `pkg_config` is the only one that knows which modules are being fetched under the hood (managing cross sysroots, `PKG_CONFIG_PATH`, etc.), and there are no ways for this PoC access just the `run` function and then parse the output manually; it's why I added a step that reconstitutes them.

PR #477, opened 2025-07-23
Fix non-POSIX paths in Libdir under Windows

amyspark:
Hi @lu-zero ,

This fixes using the GStreamer plugins .pc files with e.g. CMake, which interprets backslashes as an escape code and strips them automatically.

This is an upstream fix for the following GStreamer MR: https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/1895

cc @nirbheek

### DIFF
--- a/src/pkg_config_gen.rs
+++ b/src/pkg_config_gen.rs
@@ -111,7 +111,7 @@ impl PkgConfig {
         }
 
         let libs = vec![
-            format!("-L{}", libdir.display()),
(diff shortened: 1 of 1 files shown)


nirbheek (approved):

lu-zero:
Sounds good, can we wait till the next cargo release or do you prefer to have a point release with this fix sooner?

nirbheek:
It can wait, we've already patched it in cerbero.

PR #483, opened 2025-08-08
Bump to cargo 0.90

lu-zero:


sdroege (approved):
Looks correct to me and works as good as before (i.e. integration tests don't work, creating an issue now).

PR #486, opened 2025-08-09
Support integration tests

lu-zero:


sdroege:
Doesn't solve it :)

sdroege:
This works now. Thanks!

PR #500, opened 2025-11-03
Add Cygwin support

lazka:
Things are like with Windows+gnu, except the "meson-paths" logic is always used, and shared libs are prefixed with "cyg".

### DIFF
--- a/.github/workflows/example-project.yml
+++ b/.github/workflows/example-project.yml
@@ -132,3 +132,45 @@ jobs:
           meson setup build
           meson compile -C build
           meson test -C build
+
+  example-project-cygwin:
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    steps:
+      - name: Clone Git repository
+        uses: actions/checkout@v5
+
+      - name: Install rust
+        uses: msys2/setup-msys2@v2
+        with:
+          update: true
+          msystem: MSYS
+          install: rust pkgconf gcc openssl-devel libopenssl make
+
+      - name: Install cargo-c applet
+        run: cargo install --path .
+
+      - name: Build C API for example project
+        working-directory: example-project
+        run: cargo cbuild --verbose --release
+
+      - name: Install into /usr/local
+        working-directory: example-project
+        run: cargo cinstall --verbose --release --prefix=/usr/local
+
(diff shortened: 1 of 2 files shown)


src/build_targets.rs
```diff
@@ -131,7 +131,9 @@ impl BuildTargets {
     pub fn shared_output_file_name(&self) -> Option<OsString> {
         match self.lib_type() {
             LibType::Windows => {
-                if self.shared_lib.is_some()
+                if self.target.os == "cygwin" {
+                    Some(format!("cyg{}.dll", self.name).into())
```
lu-zero: is this really correct?

src/build_targets.rs
```diff
@@ -326,4 +328,27 @@ mod test {
...
+            os: String::from("cygwin"),
+            env: String::from(""),
+            cfg: Vec::new(),
+            target: None,
+        };
+        let file_names = FileNames::from_target(&target, "ferris", Path::new("/foo/bar"), false);
+
+        let expected = FileNames {
+            static_lib: PathBuf::from("/foo/bar/libferris.a"),
+            shared_lib: PathBuf::from("/foo/bar/ferris.dll"),
```
nirbheek:
```suggestion
            shared_lib: PathBuf::from("/foo/bar/cygferris.dll"),
```

No?

src/build_targets.rs
lazka: Here as an example: https://cygwin.com/packages/x86_64/libglib2.0_0/libglib2.0_0-2.86.0-1

lu-zero:
@lazka also can you reset the PR with the fix folded in?

src/build_targets.rs
lazka: As far as I understood it that's the filename that rustc creates, before it gets copied to the right name via cinstall, and I don't know where I could change that (??). Same as with mingw currently.

So this breaks things.

lazka:
I added some basic tests (see commit for details). I went for separate jobs as mixing in even more conditions seemed hard to follow, but I'd be happy to give it a try though if wanted.

amyspark (approved):
LGTM

nirbheek:
I guess we need an integration test that will check the names of the on-disk files after installation.

PR #518, opened 2026-01-23
target: emit correct lib dir when cross compiling on Debian

Fabian-Gruenbichler:
cross-compiling requires passing an explicit target. for Debian, we still want to emit the Multi-Arch compatible lib dir in such a case, provided the target override matches the (cross-) target from the env.

this change is required in order for the Debian Rust packaging helpers to be able to always inject `--target`, for native and cross builds. we already do so for regular cargo invocations, but not for cargo-c ones at the moment.

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -161,11 +161,20 @@ impl Target {
     }
 
     pub fn default_libdir(&self) -> PathBuf {
-        if self.is_target_overridden || self.is_freebsd() {
+        if self.is_freebsd() {
             return "lib".into();
         }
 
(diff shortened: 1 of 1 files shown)


lu-zero (approved):
It looks fine. You have the perfect timing since tomorrow I'll update cargo-c.

Fabian-Gruenbichler:
nice, thanks :) I implemented the changes in our cargo wrapper script so that they don't change anything for native builds, but once the updated version lands in Debian, cross builds should work as well! :fireworks:

Issue #516, opened 2026-01-21
Add --path to cargo cinstall


predators46:
hy @lu-zero 

It seems like --path is missing in cargo-c could you add it? this is useful when building on openwrt
```
error: unexpected argument '--path' found

  tip: a similar argument exists: '--manifest-path'
```

lu-zero:
For cargo parity sure, but how it would be different from passing `--manifest-path {the-path}/Cargo.toml` ?

predators46:
work

```
root@OpenWrt:~# ldd /usr/lib/librav1e.so.0.8
        ldd (0x7f98150000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7f97f2e000)
        libc.so => ldd (0x7f98150000)
root@OpenWrt:~# ldd /usr/lib/librav1e.so
        ldd (0x7fb04d0000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7fb02ae000)
        libc.so => ldd (0x7fb04d0000)
root@OpenWrt:~# ldd /usr/lib/librav1e.so.0.8.1
        ldd (0x7fa72e0000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7fa70be000)
        libc.so => ldd (0x7fa72e0000)
root@OpenWrt:~# ldd /usr/bin/rav1e
        /lib/ld-musl-aarch64.so.1 (0x7f9fd40000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x7f9fd0f000)
        libc.so => /lib/ld-musl-aarch64.so.1 (0x7f9fd40000)
root@OpenWrt:~# rav1e -V
rav1e 0.8.1 () (release)

```

lu-zero:
You aren't telling me what is the command you are using and what is the outcome you'd expect...

predators46:
@lu-zero 

using --manifest-path. instead of --path. no other issues that I found

you need to set --target otherwise x86 will be selected

```
root@OpenWrt:~# rav1e --version
rav1e 0.8.1 () (release)
rustc 1.90.0 (1159e78c4 2025-09-14) (built from a source tarball) aarch64-unknown-linux-musl
Compiled CPU Features: neon
Runtime Assembly Support: Enabled
Runtime Assembly Level: NEON
Threading: Enabled
Unstable Features: Disabled
[...]

PR #519, opened 2026-01-24
Bump cargo version

lu-zero:


lu-zero:
It fixes as well https://github.com/lu-zero/cargo-c/issues/516

PR #537, opened 2026-04-21
target: add -Zunstable-options when querying cfg

DeepeshWR:
Since Rust 1.95,invoking rustc with `--target` requires `-Zunstable-options` when used with custom JSON targets. Without this flag, the command fails.

Reference:
https://github.com/rust-lang/rust/pull/152677/commits/b4e645fe5af113f6ec072e04d55b7eb742e236ca https://github.com/rust-lang/rust/pull/154993/commits/85633125c7616d00e3c3a33eb356e26540b8a438

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,3 +69,6 @@ inherits = "release-strip-lto"
 # To avoid #414
 [workspace]
 exclude = ["example-project", "example-workspace"]
+
+[dev-dependencies]
+tempfile = "3.27.0"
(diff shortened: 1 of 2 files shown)


lu-zero:
Can you please also add a test for it?

lu-zero:
adding it unconditionally breaks everything though :)

lu-zero:
Please make rustfmt and clippy happy

lu-zero:
it is still unhappy :/

DeepeshWR:
> it is still unhappy :/

Fixed now :)

lu-zero:
sadly not yet...

